Issue #19, opened 2022-05-10
Add a YAML file for creating the dev-environment?


HTenkanen:
@christophfink: Now after starting to play around with this, I notice that we don't yet have a environment file that could be used to easily create a dev-environment for the package. Shall I write one, now as I'm setting things up?

HTenkanen:
We could follow a similar style doing these as many projects, such as [geopandas](https://github.com/geopandas/geopandas/tree/main/ci/envs), i.e. there woud be a folder `ci/envs` and we could add our environments there.

christophfink:
I’m feeling a bit ignorant, but I always thought the requirements.txt/setup.cfg/pyproject.toml already describe all necessary steps? But yes, go ahead. This is (probably?) also important for a matrix testing setup that would be nice (with all the Java stuff coming in, too, and different OS and such)

christophfink:
So environment files are apparently a conda thing, so this should go together with issue #13

christophfink:
Should we close this issue now, and re-open it (or rather new ones), when we need other environments?

christophfink:
Reopening this, we should add [nbstripout](https://anaconda.org/conda-forge/nbstripout) to the developer env.

HTenkanen:
@christophfink Will also add OpenJDK here as a dependency so when developing, we can be sure that our environments share the same version of JDK: https://anaconda.org/conda-forge/openjdk

HTenkanen:
Using conda to install JDK is also extremely convenient so when we instruct how to install r5py it's basically just a couple of steps, 1) install openjdk and 2) install r5py.

christophfink:
perfect!

christophfink:
The GitHub runners only allow certain flavours of JDKs, but the default one seems to work well.

christophfink:
Also make sure these are available:
```
ImportError: The 'folium', 'matplotlib' and 'mapclassify' packages are required for 'explore()'. You can install them using 'conda install -c conda-forge folium matplotlib mapclassify' or 'pip install folium matplotlib mapclassify'.
```

HTenkanen:
Ah right, will add them.

christophfink:
Thanks!

HTenkanen:
> Let’s also make sure these are available:
> 
> ```
[...]

PR #27, opened 2022-05-10
conda environment yml file (issue #19)

christophfink:
Not sure what is an appropriate environment name (module name sounds okay)

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,16 @@
+name: r5py
+
+channels:
+    - conda-forge
+
+dependencies:
+    - python
+    - configargparse
+    - geopandas
+    - joblib
+    - jpype1
+    - numpy
+    - pandas
+    - psutil
+    - requests
+    - shapely


environment.yml
```diff
@@ -0,0 +1,16 @@
+name: r5py
+
+channels:
+    - conda-forge
+
+dependencies:
+    - python
```
HTenkanen, line 7: Typically, you would want to test against specific versions of Python, i.e. here we would specify e.g. `python=3.9` or `python=3.10`. I would aim to put a bit of control here and also be specific that the library is tested against Python versions 3.8 upwards or something. I would aim to support a bit older versions of Python (i.e. up from 3.8) because people tend to upgrade their environments a bit slowly.

HTenkanen:
In addition to comment above, I would create a folder for these environments because we are likely to have multiple .yaml files for the project.

environment.yml
christophfink: How would I specify more than one version here? (never used conda, can’t find anything on the quick)

christophfink: I would anyway carry out tests in a matrix (still have to look into how github workflows look like for that), so the multiple versions get tested anyway - so should we only specify the minimum version r5py requires (so `python>=3.8`)?

HTenkanen: Basically, you want to create a different environment file for different versions. I am just now creating one environment for Python 10 with developer tools included as well. When writing the GA workflow, there we will define the matrix for testing the package with all different environments. Take a look e.g. how geopandas does it here: https://github.com/geopandas/geopandas/blob/main/.github/workflows/tests.yaml

HTenkanen: @christophfink Now created a PR for a developer environment file in #28

HTenkanen: Named the environment as `python_310_dev.yaml`, so it's easy to understand from the file name that this env is for Python 3.10 with dev tools

christophfink: Anyway, I think, this is something you should take over entirely, since I don’t use conda. I’ll close this pull request, as it is redundant, now.

PR #28, opened 2022-05-10
Add a development environment for Python 3.10

HTenkanen:
Relates to #19


HTenkanen:
> So `codecov`, `requests`, `black` are things that you like to have in your dev environment? I’m fine with them (on a related note: should we reject pushes that don’t pass black?), just checking what’s your reasoning for them?

Yes, these are what I think we all should have in our dev environment. In an optimal scenario, one would always remember to check with black that the code style is okay before pushing / making a PR. This is "optimal scenario", because I for example tend to forget this quite often 😄 

Also one can check with `pytest` the coverage already when writing the tests, so it's much easier to find places that are not tested yet. With following command (requires the `codecov` package), you get an output showing also the test coverage file by file and line by line:
```
# inside root folder
$ pytest --cov=src/r5py -v tests/
```

As a result, you get something like this (as we see, there is a bit of work still 😆 )
image

christophfink:
Let’s just leave it in for the moment, it’s never bad to have it I guess :)

christophfink (approved):

PR #30, opened 2022-05-10
coverage testing, matrix testing

christophfink:
Work in progress (finishing my work day), but please comment, already.

This still needs a bit of work to produce a coverage badge that can be displayed in the README.md
On top of that, putting the pytest options into pyproject.toml might be smart; and the `coverage` module is worth a look.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,2 @@
+[report]
+show_missing = True
--- a/.github/workflows/lint-test-upon-push.yml
+++ b/.github/workflows/lint-test-upon-push.yml
@@ -1,12 +1,17 @@
 # vim: set tabstop=2 softtabstop=2 shiftwidth=2 expandtab:
 
-name: Lint and test after every push
-on: 
+name: Lint and test after every push or pull request
+on:
   push:
     paths:
       - "*.py"
       - "**/*.py"
+  pull_request:
   workflow_call:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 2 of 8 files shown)


HTenkanen:
To be honest, I have never used tox for anything. I have found it somehow confusing to use. Not saying that we shouldn't use it, but I'm not a correct person to review those things as I'm not familiar with it.. 

I have used `.coveragerc` [configuration file](https://github.com/HTenkanen/pyrosm/blob/master/.coveragerc) and a codecov action which is part of the GA workflow in [tests.yaml](https://github.com/HTenkanen/pyrosm/blob/master/.github/workflows/tests.yaml)

HTenkanen:
Here is some discussion about tox: https://discourse.jupyter.org/t/thoughts-on-using-tox/3497/10

christophfink:
conftest.py is not a test, but a config file (used by many test frameworks, fixed filename)

christophfink:
Going to read the discussion on tox tomorrow morning

HTenkanen:
> conftest.py is not a test, but a config file (used by many test frameworks, fixed filename)

Ah, right 👍🏻

christophfink:
@HTenkanen : Could you check whether this looks alright now?

christophfink:
Is there any good mechanism to test Github Actions (except trial-and-error)?

HTenkanen (approved):
Looks good to me!

HTenkanen:
Oof, these GAs are indeed quite annoying to debug..

Issue #29, opened 2022-05-10
When fetching R5 jar, notify user if it was a success.


HTenkanen:
A small thing, but I think it would be intuitive if we would notify the user if the downloading process of R5 jar was a success? Currently, I see this:
![image](https://user-images.githubusercontent.com/6659534/167688652-a2fbfa76-d54b-4250-bac1-e2554ce2fcde.png)

I.e. it is not necessarily very clear whether everything is now okay or not.

PR #37, opened 2022-05-11
success message, r5 jar download (issue #29)

christophfink:
Now we tell the user that we successfully downloaded an R5 jar.


christophfink:
addresses issue #29

HTenkanen (approved):
Looks good!

PR #40, opened 2022-05-11
linted with black

christophfink:
In order to force linting with black, I ran the existing code through it. There are a few lines in which I was happier with manual formatting, but overall, of course, automated formatting has tremendous advantages.

So this is literally only running `black src/`


HTenkanen (approved):
Looks good. Linting indeed in some rare cases makes the code a bit less readable, but the plus-side of having systematic code style and automatic way to check them are way more important/beneficial, imo.

Issue #36, opened 2022-05-11
Add a shorthand to load all files in a directory into a TransportNetwork


christophfink:
> 3\. Building the network
>         I believe `TransportNetwork()` and `setup_r5()` would be the equivalent function used to create a network, right?
> 
> 
>     * r5py: `network = TransportNetwork(osm_pbf=osm_pbf, gtfs=gtfs)`
> 
>     * r5r: `network =  setup_r5( data_path = "local_dir")`
> 
> 
> I see that users would need to pass the the addresses of a `osm.pbf` and `gtfs.zip` files to `TransportNetwork()`. I find this behavior a bit strange. How does it work when users have two or more GTFS feeds from different agencies. We know R5 automatically detects and takes in one `osm.pbf` and all the `gtfs.zip` files in the path. This one of the reasons we require `r5r` users to input the address of the local directory where the data is stored. Perhaps we should have a separate issue to discuss the workflow we would expect from users and the expected behavior of these download and build network functions.

This is splitting up issue #25 into smaller pieces (re: https://github.com/r5py/r5py/issues/25#issuecomment-1122901350 )

christophfink:
Honestly, this is something that low-key bothered me about r5r. We actually want specific files, not ’everything I have’. Many people like to keep all their data somewhere in a common place (thinking about historical gtfs or network data, for instance), and not in a project-specific directory.

On a side note, `gtfs` is a list of files (can be zero or more).

I could imagine building a wrapper around it, which searches for all possible GTFS files and the first (?) `.osm.pbf` to hand to the TransportNetwork constructor.

HTenkanen:
Yeah, I think the convention that is used by r5r is originating how also OpenTripPlanner works, i.e. probably follows the conventions of r5. I do agree that the logic is not perhaps the best one (at least when working with OTP, you tend to create multiple copies of the same files because of this). Being able to list specific GTFS feeds with a list would be more specific and it would allow you to have as many GTFS feeds in your directory as you want, but still you could control which of them are used. And as @christophfink mentions, it's easy to build a wrapper that searches all the feed zip files and uses them if needed. 

But again, this is a question of how much do we want to follow the original implementations and how they work (OTP + r5 + r5r). For someone, who has used these tools before (OTP, r5, r5r), the original logic might be easier to grasp, but for new users having more control might be a good feature.

HTenkanen:
Yep, my vote goes to have a method for doing this. 

[...]

PR #45, opened 2022-05-11
add TransportNetwork.from_directory() factory function (issue #36)

christophfink:


src/r5py/r5/transport_network.py
```diff
@@ -62,6 +64,47 @@ def __init__(self, osm_pbf, gtfs=[], build_config={}):
...
+        path = pathlib.Path(path)
+        try:
+            osm_pbf = sorted(path.glob("*.osm.pbf"))[0]
+        except KeyError:
+            raise FileNotFoundError(
+                "Could not find any OpenStreetMap extract file (`.osm.pbf`) in {}".format(
+                    str(path.absolute())
+                )
+            )
+        gtfs = list(path.glob("*GTFS.zip")) + list(path.glob("GTFS*.zip"))
```
HTenkanen, line 99: I guess this is case sensitive, or not? If yes, then probably would be good to deal with files that have possibly been written in different ways (lowercase, uppercase, capital)

src/r5py/r5/transport_network.py
```diff
@@ -62,6 +64,47 @@ def __init__(self, osm_pbf, gtfs=[], build_config={}):
         )
         self._transport_network.transitLayer.buildDistanceTables(None)
 
+    @classmethod
+    def from_directory(cls, path):
+        """
+        Load a transport network, find input data in `path`.
```
HTenkanen, line 70: Or "Find input data in `path` to load a transport network for r5py." ?

src/r5py/r5/transport_network.py
```diff
@@ -62,6 +64,47 @@ def __init__(self, osm_pbf, gtfs=[], build_config={}):
...
+            directory path in which to search for GTFS and .osm.pbf files
+
+        Returns:
+        --------
+        TransportNetwork
+            A fully initialised r5py.TransportNetwork
+        """
+        path = pathlib.Path(path)
+        try:
+            osm_pbf = sorted(path.glob("*.osm.pbf"))[0]
```
HTenkanen, line 92: We don't need to do anything for this now, but perhaps would be good to add a TODO comment here, so that we deal with this somehow in the future. Now we just assume to use the first osm dump that is found, but if there are more of them, I think we should at least warn the user, perhaps even raise an Exception because an incorrect OSM pbf matched with the transit network can cause unexpected behavarior.

src/r5py/r5/transport_network.py
```diff
@@ -62,6 +64,47 @@ def __init__(self, osm_pbf, gtfs=[], build_config={}):
...
 
+    @classmethod
+    def from_directory(cls, path):
+        """
+        Load a transport network, find input data in `path`.
+
+        This mimicks r5r’s behaviour to accept a directory path
+        as the only input to `setup_r5()`.
+
+        If more than one OpenStreetMap extract (`.osm.pbf`) is
```
HTenkanen, line 75: In addition, to this documentation, I think it would be good to at least warn a user if more than one PBF files were detected.

HTenkanen:
Hi @christophfink looks good! I added a few comments. Nothing critical, but something to take into consideration 👍🏻

src/r5py/r5/transport_network.py
christophfink: That’s a good point and also a mental TODO list item of mine which I forgot about: I’m still going to check which files R5 and r5r consider GTFS data sets.

christophfink: Okay, so R5 runs a slightly more sophisticated check, and I’m going to implement something along the same lines: 

1. It checks for the file extension, 
2. if it’s `.zip`, check its content
3. if it contains `stop_times.txt`, consider it GTFS

https://github.com/conveyal/r5/blob/dev/src/main/java/com/conveyal/r5/transit/TransportNetwork.java#L246

This is still a bit of a heuristic approach, but I guess we would fail later, in case there’s a non-GTFS zip file containing `stop_times.txt`. The [reference standard](https://developers.google.com/transit/gtfs/reference#field_definitions) lists `agency.txt`, `stops.txt`, `routes.txt`, `trips.txt`, `stop_times.txt` as *required*. 

This opens up yet another question (but that might be an issue for the future): should we even go there, or rather leave this all to R5 by calling its `TransportNetwork.fromDirectory()` under the hood? The thinner `r5py` stays, the easier it will be to maintain, and it would help avoid redundant code.

src/r5py/r5/transport_network.py
christophfink: I would maybe leave the r5py out there, or make it ’... load an `r5py.TransportNetwork`’ - does that sound okay?

src/r5py/r5/transport_network.py
HTenkanen: Hmm, good point. My initial opinion would be to leave this logic to R5 for obvious reasons that you said here. If  needed for some customization, we can always decide to implement this ourselves later.

src/r5py/r5/transport_network.py
HTenkanen: I think that sounds good and explicit 👍🏻

src/r5py/r5/transport_network.py
christophfink: I added it for now, because I think it makes sense, but let’s keep it in mind that this is not set in stone.

HTenkanen:
@christophfink Is this now ready for review?

christophfink:
@HTenkanen yes, just still waited for the checks to pass

src/r5py/r5/transport_network.py
```diff
@@ -62,6 +65,60 @@ def __init__(self, osm_pbf, gtfs=[], build_config={}):
...
+        try:
+            potential_osm_pbf_files = sorted(path.glob("*.osm.pbf"))
+            osm_pbf = potential_osm_pbf_files[0]
+            if len(potential_osm_pbf_files > 1):
+                warnings.warn(
+                    (
+                        "Found more than one OpenStreetMap extract file (`.osm.pbf`), "
+                        + "using alphabetically first one ({})"
+                    ).format(osm_pbf.name),
+                    warnings.RuntimeWarning,
```
HTenkanen, line 101: Minor comment, but I would recommend using f-strings instead of `.format()` because the readability of f-strings is much better. I.e. this would become:
```python
f"Found more than one OpenStreetMap extract file (`.osm.pbf`), "
f"using alphabetically first one {osm_pbf.name}."
```
Also notice that you don't need to add the `+` character here to combine multiple text lines.

HTenkanen (approved):
@christophfink Looks good! I added one minor comment about using f-strings but other than that looks great! 💪🏻

src/r5py/r5/transport_network.py
christophfink: Yes, these are only str, so yes, I can change them to f-strings

HTenkanen:
@christophfink : Did still some minor fixes, now should be good 👍🏻

HTenkanen (approved):

HTenkanen:
@christophfink : Sorry I did some direct changes without properly notifying. I think your push changed back one of my changes on [these lines](https://github.com/r5py/r5py/blob/a77399fa2b11f84647461651f3b63f03894d75b4/src/r5py/r5/transport_network.py#L96-L101) 

I.e. I believe:
```python
warnings.warn(
                    (
                        "Found more than one OpenStreetMap extract file (`.osm.pbf`), "
                        f"using alphabetically first one ({osm_pbf.name})"
                    ),
                    warnings.RuntimeWarning,
                )
```

should be:

```python
warnings.warn(
                    f"Found more than one OpenStreetMap extract file (`.osm.pbf`), "
                    f"using alphabetically first one ({osm_pbf.name})",
                    RuntimeWarning,
                )
```

With the first code, I at least get an error on my computer:

image

christophfink:
Oh, true! RuntimeWarning is a global!

christophfink:
We had some merge conflicts there - let me make that chance, still :)

HTenkanen:
Yes go ahead. 👍🏻 I thought that I make a quick fix without you needing to act on it, but here we see what happens when you do something "quick" 😆

christophfink:
> Yes go ahead. 👍🏻 I thought that I make a quick fix without you needing to act on it, but here we see what happens when you do something "quick" laughing

I thought the very same of quickly implementing that small change :D

christophfink:
Let’s squash this PR, it’s had quite a lot of commits in the end

HTenkanen:
Yes, let's do that 👍🏻

HTenkanen:
Finally there 🎉 

Resolves #45

Issue #52, opened 2022-05-11
Remove unnecessary dependencies


HTenkanen:
In our environment files and also setup.cfg, there are multiple dependencies that don't need to be there, i.e. numpy, pandas, shapely.
All of those packages are "hard" dependencies of geopandas and will be automatically installed with geopandas.

HTenkanen:
Yes well they might change but it is highly unlikely in this case, so I wouldn't be so strict with this. Geopandas is absolutely never going to drop the dependency from pandas (it's totally built on top of it) and similar case with shapely (everything relies on it). I also don't see that pandas would drop numpy as a dependency anytime soon. 

And after all, if some of these hard dependencies would be dropped (very unlikely), we can then just add the dependency to our environment. If we would rely on specific version of these underlying dependencies of geopandas, then it would be natural to have them here, but that is not the case here. I just see these redundant, and for all these environments there are unnecessary calls and steps done when the environments are created.

christophfink:
But why remove them now?

christophfink:
But the packages should be pulled in explicitly, since we actually use them, also. It's of course a bit of an academic discussion, because in practice the outcome is very similar (also dependency resolving should be very similar in effort). Still, I would assume it best practice to require all dependencies that we import. Of course not dependencies of dependencies (if we don't import them ourselves)

HTenkanen:
Okay, fair enough 👍🏻 But we do not actually import Shapely anywhere in the package (?) , so with this logic I think it shouldn't be in the setup.cfg? We only import it in documentation and I wouldn't count those dependencies in a similar manner as dependencies that the package requires.

christophfink:
I think so, I'll check in the morning

christophfink:
I'll properly grep all import statements and compare it against the requirements

christophfink:
```bash
> grep -hE '^import | from .* import' -r src/ | sed 's/^import //;s/^from \(.*\) import.*$/\1/' | sort | uniq | grep -vE '^\.'
ch.qos.logback.classic
com.conveyal.r5
configargparse
datetime
enum
hashlib
java.io
java.lang
java.time
java.util.ArrayList
joblib
jpype
jpype.imports
jpype.types
json
math
multiprocessing
numpy
org.slf4j.LoggerFactory
os
os.path
pandas
psutil
re
requests
sys
```

You were right, we don't import shapely! :smile:

christophfink:
configargparse
numpy
pandas
psutil
requests

-- 

[...]

PR #56, opened 2022-05-11
clean requirements (issue #52)

christophfink:
This addresses issue #52 - I cleaned the requirements.txt and environment.yml files, as well as the requirements of the documentation, to only depend on packages that we actually import. This meant taking geopandas and shapely off the general dependency lists, but adding a few packages to the dependencies of the documentation notebooks. 

(Note for future selves: if we want to introduce typehints at some point, we will need geopandas and shapely, again)

### DIFF
--- a/ci/python_310.yaml
+++ b/ci/python_310.yaml
@@ -6,11 +6,9 @@ channels:
 dependencies:
     - python=3.10
     - configargparse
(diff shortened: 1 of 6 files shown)


HTenkanen (approved):
Looking good! 👍🏻

PR #61, opened 2022-05-12
os.path -> pathlib

### LINKED ISSUE #58: Replace os.path with pathlib throughout
Python’s `pathlib` (builtin) library supersedes the now rather legacy `os.path`. R5py code uses both of them in a non-systematic fashion, all of the occurrences of os.path could easily be replaced by the more modern pathlib.

christophfink:
closes #58 

still waiting for pull request #45 to be merged (likely merge conflicts to be cleaned)


src/r5py/r5/transport_network.py
```diff
@@ -110,7 +92,7 @@ def from_directory(cls, path):
             if util.contains_gtfs_data(potential_gtfs_file)
         ]
         try:
-            with open(os.path.join(str(path), "build.json")) as build_json_file:
+            with open(pathlib.Path(path) / "build.json") as build_json_file:
```
HTenkanen, line 95: Does using `/` for sure work on Windows? (historically creating paths just by adding `/` has been quite problematic).

Not sure how pathlib works because I haven't used it that much, so does it take care of that the filepaths work appropriately both in Unix + Windows? I have used `os.path.join()` function to tackle these things, but that might be "old style" and I'm just not aware of how it should be done nowadays.  

This comment applies to all places where this code style is used.

src/r5py/util/classpath.py
```diff
@@ -3,7 +3,8 @@
 """Make sure R5 is in the class path, download it if not."""
 
 import hashlib
-import os.path
+import os
+import pathlib
```
HTenkanen, line 7: should we import this like `from pathlib import Path` ? It would make the code a bit cleaner, imo.

src/r5py/r5/transport_network.py
```diff
@@ -5,7 +5,6 @@
 
 
 import json
-import os.path
 import pathlib
```
HTenkanen, line 8: Import as `from pathlib import Path` ?

HTenkanen (commented):
Checked the codes and added a couple of comments / questions.

src/r5py/r5/transport_network.py
christophfink: that’s why pathlib - it standardises those things, and `/` is in fact an operator here, and not a character in the path

src/r5py/r5/transport_network.py
christophfink: I know many people like to do these abbreviated imports, and also numpy as np and pandas as pd, but I always feel uneasy about those things. Code should be as explicit as possible, so I don’t have to go to the top of the file to figure out where something came from. My personal convention there is to only do absolute imports, except for resources from within the same package.

That said, of course you can outvote me on that, and I’ll accept it ;)

HTenkanen: Well I guess this a bit of a question of taste / what one is used to do, and there are two different schools to this as you mention. Personally, I prefer having more clean / shorter code which to me is more readable, and that can be achieved with abbreviated imports. However, I'm okay going either way, as long as the style is systematic. So you can decide :)

HTenkanen: Yes, with the java paths it definitely makes sense to be as explicit as possible. So let's aim to follow the same style throughout the project.

src/r5py/r5/transport_network.py
HTenkanen: Right! Good to update my old habits here also 🙂 Need to start using pathlib more.

HTenkanen:
@christophfink Is this still a WIP or can this be merged?

christophfink:
Still WIP, found some more in docs (and that now overlaps with working on the issue with the Java exception

christophfink:
Now removed `myst_nb` for a moment, because it’s what caused recursion and troubles on my local sphinx build, let’s see what RTD says.

christophfink:
RTD runs into this one: https://github.com/r5py/r5py/issues/19#issuecomment-1124949763 

> Let’s also make sure these are available:
> 
> ```
> ImportError: The 'folium', 'matplotlib' and 'mapclassify' packages are required for 'explore()'. You can install them using 'conda install -c conda-forge folium matplotlib mapclassify' or 'pip install folium matplotlib mapclassify'.
> ```

christophfink:
It works with `nbsphinx` instead of `myst_nb`. Can we go forward with this as-is, and add `myst_nb` in the next version?

christophfink:
There remain a few places (also in this PR), in which f-strings would be nicer than str.format(), but let’s make that a separate issue and effort.

HTenkanen (approved):
Looks good for now!

PR #62, opened 2022-05-12
Proposed license change (MIT in addition to GPLv3)

christophfink:


HTenkanen:
Here is some discussion why and how dual-license work:
https://softwareengineering.stackexchange.com/questions/139663/confusion-about-dual-license-mit-gpl-javascript-for-use-on-my-website

HTenkanen (approved):
Looks great!

HTenkanen:
@christophfink : By the way, not sure if you already know this, but if you add `[skip ci]` to commit message, you can skip the GA pipeline. For example, in this case we wouldn't have needed to run the tests.

christophfink:
That's a great tip that I did not know! Thanks!

PR #68, opened 2022-05-12
Add issue templates

christophfink:


christophfink:
These are issue templates that GitHub suggests, plus a bit of fiddling to fit them a bit better to Python in general, and to r5py, in particular.

PR #75, opened 2022-05-12
f-strings instead of str.format()

### LINKED ISSUE #69: Replace str.format() with f-strings where possible
re: https://github.com/r5py/r5py/pull/61#issuecomment-1125052362

christophfink:
Replaced all occurrences of str.format() with the faster, neater, and shorter f-strings.

Closes #69

### DIFF
--- a/src/r5py/util/validatingrequestssession.py
+++ b/src/r5py/util/validatingrequestssession.py
@@ -54,9 +54,7 @@ class ValidatingRequestsSession(requests.Session):
 
         if digest != checksum:
             raise ChecksumFailed(
-                ("Checksum failed for {}" + "expected {}, got {}").format(
-                    url, checksum, digest
-                )
+                f"Checksum failed for {url}" + "expected {checksum}, got {digest}"
             )
 
         return response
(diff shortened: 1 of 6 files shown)


src/r5py/util/validatingrequestssession.py
```diff
@@ -54,9 +54,7 @@ def request(self, method, url, checksum, **kwargs):
 
         if digest != checksum:
             raise ChecksumFailed(
-                ("Checksum failed for {}" + "expected {}, got {}").format(
-                    url, checksum, digest
-                )
+                f"Checksum failed for {url}" + "expected {checksum}, got {digest}"
             )
```
HTenkanen, line 58: This won't produce expected text. The correct syntax would be:
```python
f"Checksum failed for {url}, expected {checksum}, got {digest}."
```

HTenkanen (changes requested):
Looks good! Found one f-string that doesn't work as expected, added suggestion for the correct syntax.

src/r5py/util/validatingrequestssession.py
christophfink: Good spot! How did I miss this?

HTenkanen (approved):
Looks good!

PR #76, opened 2022-05-12
additional unit tests

christophfink:
Trying to catch up with our backlog of missing unit tests


christophfink:
Let’s add a few tests (11 fresh from today, a few more from last week). Merging this to main once in a while is a good idea, I guess.

tests/test_transport_network.py
```diff
@@ -0,0 +1,113 @@
...
+        return r5py.TransportNetwork.from_directory(DATA_DIRECTORY)
+
+    def test_init_from_files_and_dir_cover_same_extent(
+        self, transport_network_from_test_files, transport_network_from_test_directory
+    ):
+        assert (
+            transport_network_from_test_files._transport_network.getEnvelope()
+            == transport_network_from_test_directory._transport_network.getEnvelope()
+        )
+        # assert (
```
HTenkanen, line 37: @christophfink Are these comments important? Wouldn't want to merge these kind of unused codeblocks to main, so perhaps clean up from here?

christophfink: I thought of leaving these here for reference, to remember that this precisely does not work, but a similar approach might be good

HTenkanen: Hmm, could you rather e.g make an issue or add a comment to current issue with these lines of codes for reminder? I have a bit annoying experiences cleaning these kind of "I just leave these here for now" kind of lines in the past 😆 We have made good efforts keeping the codebase clean, and would like to continue doing so 🙂

christophfink: Absolutely, I'll convert them into comments or an issue (let's see how I feel in the morning)

pyproject.toml
```diff
@@ -3,9 +3,10 @@ build-backend = "setuptools.build_meta"
 requires = ["setuptools>=50.0.0", "wheel"]
 
 [tool.pytest.ini_options]
-addopts = "-r s --cov=./src/r5py --cov-append --cov-report term-missing --cov-report xml"
+addopts = "-p no:faulthandler -r s --cov=./src/r5py --cov-append --cov-report term-missing --cov-report xml"
```
HTenkanen, line 6: Is `no:faulthandler` used to ignore the segfault warnings, or is it something else?

tests/test_transport_network.py
```diff
@@ -0,0 +1,98 @@
...
+    ):
+        assert (
+            transport_network_from_test_files._transport_network.getEnvelope()
+            == transport_network_from_test_directory._transport_network.getEnvelope()
+        )
+
+    @pytest.mark.parametrize(
+        ["transport_network"],
+        [
+            (pytest.lazy_fixture("transport_network_from_test_files"),),
```
HTenkanen, line 43: Question to educate myself as haven't seen these before 🙂 What are these lazy fixtures? What do they do?

HTenkanen (approved):
Looks very good and nice that you are parametrizing the tests to make them clean and concise! 💪🏻 Didn't spot any issues, but asked a couple of questions to better understand some things. 🙂

pyproject.toml
christophfink: It's needed because of how jpype is designed (replacing what was in conftest.py). If I understand correctly, the jpype superimposes the python faulthandler on top of/instead of (not sure) the JVM's. When pytest adds its own still on top of those, things stop working. 

https://github.com/jpype-project/jpype/issues/842

PR #84, opened 2022-05-13
improve basic usage tutorial

christophfink:
this is just one small step: rename variables to reflect the renaming of the referenced classes


HTenkanen (approved):
Looks good!

PR #85, opened 2022-05-13
better relative import order

christophfink:
Reorganised relative imports to be more explicit and more specific.

Also addressed the following TODO comment found in ` src/r5py/util/classpath.py`:

> TODO: print this only when --verbose is specified,
> (problem: verbosity.py already wants a jvm running)


christophfink:
I think this would be ready to go

src/r5py/r5/__init__.py
```diff
@@ -2,7 +2,8 @@
 
 """R5 classes."""
 
-from .. import util  # noqa: F401
+# start the JVM first thing
+from ..util import jvm  # noqa: F401
```
HTenkanen, line 6: These are good, but spotted that the import for `SpeedUnit` is missing and I guess the class does not exist yet in the project?

I.e. [this line](https://github.com/r5py/r5py/blob/585947f56614964089d2cf6bd89ede804d40f247/src/r5py/r5/__init__.py#L22)

src/r5py/r5/speed_config.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, **kwargs):
             See https://github.com/conveyal/r5/blob/v6.6/src/main/java/com/conveyal/r5/point_to_point/builder/SpeedConfig.java#L10
         """
         super().__init__()
-        kwargs = {util.snake_to_camel_case(key): value for key, value in kwargs}
```
HTenkanen, line 47: Looks good, but as a side note: Isn't this convention now going against what we discussed earlier here?:
https://github.com/r5py/r5py/pull/61#discussion_r871329494

HTenkanen (commented):
Looks good. Found one missing class (?) and then a question/comment about import conventions. Do we want to revisit or fine down the conventions related to imports?

src/r5py/r5/speed_config.py
christophfink: It’s a local import, so it goes with my understanding of the convention (classes and/or functions from within the module), but yes, it’s of course worth a discussion. Should it then be `r5py.util.snake_to_camel_case()`?

src/r5py/r5/__init__.py
christophfink: True! Just have to get off my train, but I’ll check as soon as I have a chance.

christophfink: I remember that I had foreseen a `SpeedUnit` enum at some point, but then decided against it (not enough benefits for the maintenance effort). Clearing that out, I seem to have remove the wrong import here: it should have been `SpeedConfig`

src/r5py/r5/speed_config.py
christophfink: hm, thinking about it, I would not want to import `r5py.util`, as I’m afraid an existing installation might override a local directory (but that might not be true, have to research a bit on that). Can we move this into a separate issue? It would be nice to define this convention (and related ones) more systematically, explore the implications (global import will call `__init__.py` first), and apply them throughout. In general, I have the feeling that in some places (e.g. `jvm.py`), my code relies too much on side effects (see also the `noqa: F401` comments). We should maybe have a `start_jvm_if_its_not_yet_running()` function, for instance.

HTenkanen: @christophfink : Sounds good, I see you have already created an issue for this in #95 👍🏻 Let's continue there. It would indeed be good to document / agree on the used conventions, because it will also help in the future when maintaining the package and when possible new contributors come in.

HTenkanen (approved):
Looks good! 👍🏻

Issue #17, opened 2022-05-10
Add code of conduct


HTenkanen:
It might also be good to add a code of conduct in a similar manner as how geopandas does it: https://geopandas.org/en/stable/community/code_of_conduct.html

wklumpen:
I can do this - I assume it would sit in the docs?

christophfink:
There’s a location suggested by GitHub community standards: https://github.com/r5py/r5py/community

HTenkanen:
> I can do this - I assume it would sit in the docs?

@christophfink: Actually, I believe it would be good to put this document inside the `docs` directory because we want to include this document to the website as well, and Sphinx doesn't seem to support adding files that are "above" the `docs` folder (i.e. on project root).

christophfink:
Does that allow us to somehow also make GitHub see that we added a Code of Conduct? (so we get the green check mark there :D)

HTenkanen:
@christophfink : Yes, Github detects all files stored in docs folder and root (there was a third location as well but don't remember what it was)

HTenkanen:
This is now merged to main by #87

PR #87, opened 2022-05-13
Create CODE_OF_CONDUCT.md

wklumpen:
Standard language with a small tweak to be open to contributions in many forms.


wklumpen:
Also see #17

christophfink (approved):

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,129 @@
...
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our
+community include:
+
+* Demonstrating empathy and kindness toward other people
```
HTenkanen, line 20: Minor thing, but we have used `-` character elsewhere for bullet points. To be systematic, these could still be changed to follow the same convention.

HTenkanen (changes requested):
@wklumpen Looks good, small comment about bullet points. Could you change the location of the file to be inside `docs` folder? 

In addition, could you also add the code of conduct to the index.rst file (in the docs folder). In this way, we can also get this visible to our website. If you haven't worked with sphinx before, you can take a look how the CONTRIBUTING.md was added there and add this below it in a similar fashion.

CODE_OF_CONDUCT.md
wklumpen: I followed the GitHub template/convention for file location since I figured that way it would be in the standard location - but we can  move it to Docs pretty easily. Same thing with the bullet point convention though that is easily changable.

wklumpen:
All I added was `- Understanding that contributions come in many forms, and all experience levels are valuable.`

wklumpen:
@HTenkanen I changed the bullet point styling - Is it possible to relocate the file through the GH interface to avoid a pull/push?

HTenkanen:
> @HTenkanen I changed the bullet point styling - Is it possible to relocate the file through the GH interface to avoid a pull/push?

I believe it is somehow possible to do in GH directly, but possibly it's just easier to do the change locally and push again.

CODE_OF_CONDUCT.md
HTenkanen: Yep, by default GH uses root, but docs is also allowed and better here because we want the file to be visible on our docs.

wklumpen:
Turns out you just edit the file and update the path.

HTenkanen (approved):
Looks good! I still added the `CODE_OF_CONDUCT.md` to the `index.rst` file, so we can get it visible in our website.

PR #92, opened 2022-05-14
Fixed hash assertion issue on R5 pull.

### LINKED ISSUE #91: R5 Hash assertion always fails
It appears the assertion that checks the hash in `util/classpath.py` always fails and an R5 Jar is therefore always downloaded.

At first I thought it was simply that we were calling `.hexdigest` isntead of `hexdigest()` (pull request incoming for that change) but even after that fix the `AssertionError` is thrown and the file is re-downloaded.

I checked the hash strings and they are equal so I'm not sure why the `AssertionError` is being thrown at all since the strings should be equal.

```
Jar hash: 9e4ceb85a09e750f146f95d98013eb164afac2dfc900a9e68e37ae925b1ec702
Our hash: 9e4ceb85a09e750f146f95d98013eb164afac2dfc900a9e68e37ae925b1ec702
```

However pulling the string as a variable and then doing the assertion seems to work fine:

```python
with open(R5_CLASSPATH, "rb") as jar:
            our_hash = hashlib.sha256(jar.read()).hexdigest()
            assert our_hash == R5_JAR_SHA256   # Does not fail assertion
```

So I'll suggest that change.

wklumpen:
Resolves #91


HTenkanen:
Hmm, I cannot seem to be able to reproduce the issue with the (fixed) hash assertion:
image

When debugging this I noticed that if you execute the `hexdigest()` twice with the jar-file it produces a different result:
image

Could this somehow explain the weird behavior that you got @wklumpen when debugging this?

src/r5py/util/classpath.py
```diff
@@ -36,7 +36,8 @@
     R5_CLASSPATH = str(pathlib.Path(config.CACHE_DIR) / pathlib.Path(R5_JAR_URL).name)
     try:
         with open(R5_CLASSPATH, "rb") as jar:
-            assert hashlib.sha256(jar.read()).hexdigest == R5_JAR_SHA256
```
HTenkanen, line 39: I saw your comments in the issue, but I don't quite understand why following shouldn't work:
```
assert hashlib.sha256(jar.read()).hexdigest() == R5_JAR_SHA256
```
Perhaps [this comment](https://github.com/r5py/r5py/pull/92#issuecomment-1126889001) might explain the weird behavior that you experienced? Not sure, but would be very strange if this would require saving the hash into a variable for it to work.

HTenkanen (commented):
Nice job spotting the missing parantheses! However, after experimenting cannot reproduce the behavior that you had @wklumpen in terms of needing to save the hash to a variable for this to work. 🤔

christophfink:
The repeat execution difference is because the file handle now points to the end of the file. If you call jar.seek(0) in between the digest generation calls, the checksums should be the same

christophfink:
And yes, @wklumpen good spot! (That's why we should write a unit test for this :D)
I agree that simply adding the missing parentheses would be the cleanest fix

src/r5py/util/classpath.py
christophfink: I’d agree with @HTenkanen  that simply adding the parentheses would be the better option (yes, we don’t have to care about garbage collection in Python, but there’s also no reason for introducing another variable - no reuse or anything)

christophfink (changes requested):
See detailed comments :)

wklumpen:
My main issue was that adding the missing parentheses didn't help! I checked the hash manually and it was the same hash, but doing the assert in one line failed, while assigning the variable did not - I have no reasonable explanation as to why...

christophfink:
Hm, I just tried it out, and for me, it works perfectly with simply the missing parentheses added.

christophfink:
I’m wondering how to best write a unit test for this module 🤯

HTenkanen:
@christophfink @wklumpen : What is the plan with this PR? Should we just merge it and see whether it works?

christophfink:
I’d go for 

-  writing a test:
    1. import r5py, 
    2. unload the module (how?), 
    3. import again, 
    4. check whether file was downloaded a second time (maybe from the file modification timestamp?)), and 
- adding the pair of parentheses, only (and let the test check whether it works (it should))

HTenkanen:
> I’d go for
> 
> * writing a test:
>   
>   1. import r5py,
>   2. unload the module ([how?](https://docs.python.org/3/library/importlib.html#importlib.reload)),
>   3. import again,
>   4. check whether file was downloaded a second time (maybe from the file modification timestamp?)), and
> * adding the pair of parentheses, only (and let the test check whether it works (it should))

Sounds good 👍🏻 Anyway having a unit test for this is important to have. Who shall go for writing this? I'm now working on `r5py` but naturally if @wklumpen or @christophfink you want to advance with this, go ahead 🙂

christophfink:
I’m finishing up for today, as I have to be places in an hour, but I can have a go at this later tonight, or maybe over the weekend (but feel free to go ahead, @wklumpen , and let me know if you start working on this)

christophfink:
even easier:

1. import r5py.util.classpath
2. reload that module
3. check whether file was downloaded a second time.

No need to start the whole JVM and stuff.

wklumpen:
My suggestion for this PR would be to keep the unit test separate as this is a very small fix we should certainly have merged in (with the parentheses)

christophfink (approved):

PR #38, opened 2022-05-11
Add contribution guidelines

HTenkanen:
Closes #16 

Added now contribution guidelines based on how they were written for pyrosm / momepy. Updated them a bit and converted the file to Markdown. The guidelines will be also visible on the website.

PR #96, opened 2022-05-16
Add separate configuration section to documentation

### LINKED ISSUE #90: A separate section for different configuration options
@christophfink I'd say quiet output. We might actually want to document this in the docs as well. There could be a separate section about different configuration options where we collect some tips to typical aspects a user might want to configure?

_Originally posted by @HTenkanen in https://github.com/r5py/r5py/issues/88#issuecomment-1126637207_

christophfink:
Resolves #90


HTenkanen (commented):
@christophfink Should we enforce the use of Markdown instead of rst? I think most people are generally more comfortable with Markdown files and Notebooks, hence, I think it might be good to write markdown files when possible. What do you think? 

This would mean that:
- installation.rst --> installation.md
- configuration.rst --> configuration.md

Others would stay as they are.

christophfink:
I totally would support that - with rst, I have to look up things, all the time

christophfink:
- Converted *installation* and *configuration* to markdown. 
- Added a paragraph to the contribution guidelines
- Fixed a few minor things in the documentation on the way

Please check for factual correctness, but also for language issues (feel a bit tired, words are coming slowly)

why-waste-time-when-few-word-do-trick

wklumpen (approved):
I only see minor changes (1 line in each file) but they look good!

docs/conf.py
```diff
@@ -32,10 +32,13 @@
     "sphinx.ext.napoleon",
     "IPython.sphinxext.ipython_console_highlighting",
     #"myst_nb",
+    "myst_parser",
```
HTenkanen, line 35: Do not add `myst_parser` to conf.py because it causes error in Sphinx build, see commits under this comment:
https://github.com/r5py/r5py/pull/38#pullrequestreview-969167018

docs/conf.py
```diff
@@ -32,10 +32,13 @@
...
     "nbsphinx"
 
 ]
 
+source_suffix = [".rst", ".md"]
```
HTenkanen, line 40: Do not add `source_suffix` to conf.py because it causes error in Sphinx build, see commits under this comment:
https://github.com/r5py/r5py/pull/38#pullrequestreview-969167018

docs/conf.py
HTenkanen: It's weird because Sphinx documentation tells to do this, but apparently these are taken care of by `sphinx_book_theme`.

docs/index.rst
```diff
@@ -9,8 +9,9 @@ R5py
    :caption: Contents
 
    installation
+   configuration
    basic-usage.ipynb
-   CONTRIBUTING.md
+   CONTRIBUTING
```
HTenkanen, line 14: Are you sure this works without the `.md` file extension? Could well be. Mostly asking to update my knowledge if this is the case.

HTenkanen (changes requested):
Looks good! Couple of questions and one change request for the `conf.py` to avoid Sphinx build errors in RTD.

docs/conf.py
HTenkanen: Hmm, I take these comments back. It seems that `myst_nb` was taking care of these, but now as we are not using it, the `myst_parser` and `source_suffix` seem to be required.

docs/conf.py
HTenkanen: My bad. We do seem to need this because `myst_nb` is not used.

HTenkanen (approved):
Everything looks good!

PR #99, opened 2022-05-17
Update GA workflow to get test coverage properly working

### LINKED ISSUE #98: BUG: Fix GA workflow in terms of pytest coverage
Currently the pytest coverage only covers the files in `tests`.. Which is the reason for incorrect 100 % coverage badge that we have atm. 

Was debugging this for pyrosm and found a fix, so can do the same tricks for r5py as well.

HTenkanen:
Resolves #98


HTenkanen:
@christophfink : Also added the script to use mamba for installations because it is significantly faster at solving dependencies and installing packages.

christophfink:
`pip install -e` does not work with pep517/pep518 build scripts. 

However, [this](https://stackoverflow.com/a/50156706) works as expected, at least locally. I also removed the `tox.ini` and added the pytest and pytest-cov options to `pyproject.toml`. 

I’m committing the suggested changes to see whether it works in the GA environment, too.

christophfink (approved):
Since I also made changes, please take another look @HTenkanen .
Note that I moved *all* command line arguments to pytest into pyproject.toml, so they are kept in one central place.

christophfink:
coverage is now reported at a more realistic 60%: https://app.codecov.io/gh/r5py/r5py/branch/issue_98

HTenkanen:
@christophfink Thanks for the review and updates! Looks good 👍🏻 Will also update pyrosm accordingly.

PR #111, opened 2022-05-19
Arch User Repository PKGBUILD

christophfink:
This *should* now create a PKGBUILD (build recipe) to be uploaded to the Archlinux User Repository. Not quite sure still how to debug Github Actions.

### DIFF
--- a/.github/workflows/PKGBUILD.template
+++ b/.github/workflows/PKGBUILD.template
@@ -9,7 +9,7 @@ pkgver=
 pkgrel=1
 
 arch=("any")
-license=("GPL")
+license=("GPL" "MIT")
 
 makedepends=(
     "python-build"
--- a/.github/workflows/PKGBUILD.template
+++ b/.github/workflows/PKGBUILD.template
@@ -1,0 +1,42 @@
+# Maintainer: peippo <christoph+aur@christophfink.com>
+
+pkgname=python-r5py
+_name=${pkgname#python-}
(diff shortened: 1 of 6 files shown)


christophfink:
This is a bit of a troublesome PR, as for testing it, we might have to create an actual release. Maybe (maybe!), we can also trigger it manually on an existing version tag (:crossed_fingers: )

.github/workflows/PKGBUILD.template
```diff
@@ -0,0 +1,42 @@
...
+pkgname=python-r5py
+_name=${pkgname#python-}
+pkgdesc="Rapid Realistic Routing with R5 in Python"
+url="https://github.com/r5py/r5py"
+
+pkgver=
+pkgrel=1
+
+arch=("any")
+license=("GPL")
```
HTenkanen, line 12: Is it possible to have the dual license here as well?

HTenkanen (approved):
I don't really have the required expertice to properly evaluate the linux build, but looks good as far as I can understand.

.github/workflows/PKGBUILD.template
christophfink: I wasn’t sure, and then I forgot to look it up, but yes!

christophfink:
Screenshot from 2022-05-19 20-11-40

So, we’ll need a version tag to debug this. I’ll triple read over this still, and also add a debian and rpm package script, first

PR #117, opened 2022-05-20
Fix binder URL [skip ci]

### LINKED ISSUE #116: Fix Binder link in the repo Readme
Seems that an incorrect URL slipped to Readme file badge for Binder. In the Notebook itself, the Binder launcher works as expected.

HTenkanen:
Resolves #116


christophfink:
It looks like it works, but let’s wait for PR #118, maybe? It will change this path, again.

HTenkanen:
> It looks like it works, but let’s wait for PR #118, maybe? It will change this path, again.

Right, yes it does 👍🏻 Let's wait and update accordingly.

HTenkanen:
> 

Hmm, is this issue happening now in the `main` branch?

HTenkanen:
> With the new binder link, yes

Hmm, quite weird. I'm not able to reproduce this. Everything works fine on my computer (using Binder) when I just tested it. 🤔

christophfink:
> > With the new binder link, yes
> 
> Hmm, quite weird. I'm not able to reproduce this. Everything works fine on my computer (using Binder) when I just tested it. thinking

Are you maybe logged in, or so? I’m running this in an incognito window to see how a new user would experience it.

HTenkanen:
> > > With the new binder link, yes
> > 
> > 
> > Hmm, quite weird. I'm not able to reproduce this. Everything works fine on my computer (using Binder) when I just tested it. thinking
> 
> Are you maybe logged in, or so? I’m running this in an incognito window to see how a new user would experience it.

Tried this now as well. Didn't receive this error, but the memory seemed to run out of my instance when running the notebook. Will need to think about this a bit. Binder instances don't have much memory (and I think the amount of RAM might vary), so we might want to run the r5py with very low memory footprint. Alternatively, we could setup Google Colab for testing `r5py` because it has much more available memory (I think 16 GB is quite normal).

HTenkanen:
> Will need to think about this a bit. Binder instances don't have much memory (and I think the amount of RAM might vary), so we might want to run the r5py with very low memory footprint. Alternatively, we could setup Google Colab for testing `r5py` because it has much more available memory (I think 16 GB is quite normal).

Actually, I think I found a solution for this. I think we can use [GESIS Notebooks Binder](https://notebooks.gesis.org/about.html) as backend which seems to have a minimum of 4GB of memory based on their docs which should be enough for our purposes (I actually always had 8GB which is even better). This also launches much faster compared to MyBinder which is also good.

christophfink:
Great! 4GiB is plenty! It might be that we are becoming a victim of my "smart" memory management there: I think I didn't account for less than 2GiB of RAM. I'll open an issue for this, too.

christophfink:
Is this ready for review, @HTenkanen?

christophfink (approved):
Works :) :+1:

PR #121, opened 2022-05-23
run on systems with 2GiB RAM or less (issue #120)

### LINKED ISSUE #120: If less than 2GiB RAM, JVM heap size is limited to 0!
In `src/r5py/util/memory_footprint.py`, we make adjustments to the requested (or default) limits for the JVM heap size. In particular, if a relative share of total memory is requested as the set maximum (the default is 80%), and the remainder (20% using the default value) would be less than 2GiB (left for system and other services), we clamp our max-memory value to `total - 2GiB`.

This does not take into account that there might be systems that have 2GiB or less of RAM. Online notebooks, such as binder, do not provide this much. In such cases, we end up with a negative maximum heap size (does this roll over to the max value?).

I suggest to define an absolute minimum of memory that is always granted to the JVM.

~~I'll add a more detailed description once I'm back at a computer :)~~

christophfink:
closes #120

### DIFF
--- a/src/r5py/util/memory_footprint.py
+++ b/src/r5py/util/memory_footprint.py
@@ -5,11 +5,15 @@
 
 import psutil
 import re
+import warnings
 
 from . import config
 
 
 __all__ = ["MAX_JVM_MEMORY"]
+
+
+ABSOLUTE_MINIMUM_MEMORY = 200  # never grant less than 200 MiB to JVM
 
 
 config.argparser.add(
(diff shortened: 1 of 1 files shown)


christophfink:
Oh and yes, I realise that we still f*ck up systems that have less than 200 MiB RAM, but maybe that’s okay in 2022 for a package that is designed to do massive computation

HTenkanen (approved):
Looks good to me!

PR #124, opened 2022-05-23
Automatically change binder link to point to stable or main

christophfink:


christophfink:
I also created a new branch `stable` and merge `v0.0.3` into it. 
Test the binder for that version here: https://notebooks.gesis.org/binder/v2/gh/r5py/r5py/stable?urlpath=lab/tree/docs/basic-usage.ipynb

docs/conf.py
```diff
@@ -50,11 +50,22 @@
...
 html_theme = 'sphinx_book_theme'
 
+# if we are building a stable version, point the binder link to branch `stable`,
+# otherwise, point to branch `main`
+import os
+try:
+    if os.environ["READTHEDOCS_VERSION"] == "stable":
+        BINDER_BRANCH = "stable"
+    else:
+        BINDER_BRANCH = "main"
```
HTenkanen, line 60: @christophfink : I think this might be okay as it is, but just a comment: There can be situations when the docs are built in RTD for other "versions" than stable or main. E.g. at times, I have tested docs related things in RTD from branches other than main/stable. If I e.g. would build RTD docs from branch `"122-binder-stable-latest"`, the Binder link would point to "main" which differs from the one that would be expected. Is this really an issue, probably not, but we could easily fix this. If we would always fetch the branch from `os.environ["READTHEDOCS_VERSION"]`, I believe the Binder link in the docs would always be correctly pointing to documents currently being built. I.e. I believe following should work with the RTD:

```python
import os
try:
    BINDER_BRANCH = os.environ["READTHEDOCS_VERSION"]
except KeyError:  # not running on RTD 
    BINDER_BRANCH = "main" 
```

HTenkanen (approved):
This looks good! 👍🏻  One comment and possible change related to how `BINDER_BRANCH` is parsed.

docs/conf.py
christophfink: I thought about this. The problem is that in case of a PR being built, READTHEDOCS_VERSION is simply a number (e.g., "124" for PR #124). 
so we have to fall back to something sensible :/ I'll check whether we can get the related commit, maybe

HTenkanen: Hmm, are you sure this the case? In my experience, the version (at least in the online interface) has always had the name of the branch being used 🤔 Naturally if your branch name would just be a number, then this what you end up with.

christophfink: I had a test commit to check out the values of the variable. But I have another idea, give me 1/2h :)

christophfink: this should work as expected, please check again, @HTenkanen

HTenkanen: @christophfink : Excellent, this is exactly what I had in mind as well 👍🏻

HTenkanen:
@christophfink : Looks good, okay to merge. 🙂

HTenkanen (approved):

Issue #23, opened 2022-05-10
write unittests for all functions and methods


christophfink:
EDIT 2022-11-30: updated check list from coverage report, see below for original list

- r5/regional_task.py
    - [ ] 505
- r5/transit_layer.py
    - [ ] 90
- util/data_validation.py
    - [ ] 30
- util/jvm.py
    - [ ] 55
- util/memory_footprint.py
    - [ ] 33

```
- [ ] r5/regional_task.py:class RegionalTask:
- [ ] r5/regional_task.py:    def __init__(
- [ ] r5/regional_task.py:    def access_modes(self):
- [ ] r5/regional_task.py:    def access_modes(self, access_modes):
- [ ] r5/regional_task.py:    def departure(self):
- [ ] r5/regional_task.py:    def departure(self, departure):
- [ ] r5/regional_task.py:    def departure_time_window(self):
- [ ] r5/regional_task.py:    def departure_time_window(self, departure_time_window):
- [ ] r5/regional_task.py:    def destinations(self):
- [ ] r5/regional_task.py:    def destinations(self, destinations):
- [ ] r5/regional_task.py:    def egress_modes(self):
- [ ] r5/regional_task.py:    def egress_modes(self, egress_modes):
- [ ] r5/regional_task.py:    def max_bicycle_traffic_stress(self):
- [ ] r5/regional_task.py:    def max_bicycle_traffic_stress(self, max_bicycle_traffic_stress):
- [ ] r5/regional_task.py:    def max_public_transport_rides(self):
- [ ] r5/regional_task.py:    def max_public_transport_rides(self, max_public_transport_rides):
- [ ] r5/regional_task.py:    def max_time(self):
- [ ] r5/regional_task.py:    def max_time(self, max_time):
- [ ] r5/regional_task.py:    def max_time_cycling(self):
- [ ] r5/regional_task.py:    def max_time_cycling(self, max_time_cycling):
- [ ] r5/regional_task.py:    def max_time_driving(self):
- [ ] r5/regional_task.py:    def max_time_driving(self, max_time_driving):
- [ ] r5/regional_task.py:    def max_time_walking(self):
- [ ] r5/regional_task.py:    def max_time_walking(self, max_time_walking):
- [ ] r5/regional_task.py:    def percentiles(self):
- [ ] r5/regional_task.py:    def percentiles(self, percentiles):
- [ ] r5/regional_task.py:    def origin(self):
- [ ] r5/regional_task.py:    def origin(self, origin):
[...]

PR #128, opened 2022-05-26
Add tests for memory footprint

HTenkanen:
Tests related to #23


HTenkanen:
@christophfink Will finalize this tomorrow and consider the changes in #131

src/r5py/util/memory_footprint.py
```diff
@@ -113,7 +113,7 @@ def _get_max_memory(max_memory):
 
     Returns
     -------
-    float
+    int
```
HTenkanen, line 116: Changed the return to (rounded) integer because the << operator requires integer to work

src/r5py/util/memory_footprint.py
```diff
@@ -124,20 +124,20 @@ def _get_max_memory(max_memory):
     else:
         # convert to MiB
         if unit is None:
-            value *= 2**-20
+            value = value >> 20
         elif unit == "K":
-            value *= 2**-10
```
HTenkanen, line 129: These did not return correct values, so I changed the logic to use Python shift operators (<< and >>), which is also a bit cleaner code I think.

tests/test_memory_footprint.py
```diff
@@ -0,0 +1,71 @@
...
+        value = "1"
+        with pytest.warns(
+            RuntimeWarning, match="Requested maximum JVM heap size is too low for R5"
+        ):
+            assert _get_max_memory(value) == 200
+
+    @pytest.mark.parametrize(
+        ["share", "leave_at_least", "expected"],
+        [
+            (0.8, 2000, 0.8 * psutil.virtual_memory().total / (2**20)),
```
HTenkanen, line 64: Hmm, testing these are a bit tricky because the expected result seem to vary from computer to computer. On my local machine, these values produce correct results and pass the tests, but on GA these seem to fail.

HTenkanen:
@christophfink : This was a bit messy (had some issues with getting some memory tests working properly), but this is now ready for review, so could you please check 🙏🏻 Did some updates to the `memory_footprint.py` as well, and fixed some issues in converting the memory values. There were some issues and also found a cleaner way / syntax to do the conversion.

tests/test_memory_footprint.py
christophfink: maybe `virtual_memory().total` is changing on virtual containers in-between two calls?

src/r5py/util/memory_footprint.py
```diff
@@ -72,14 +72,14 @@ def _parse_max_memory_string(max_memory):
...
-        - value (float): Amount of memory to be allocated in a given unit.
+        - value (int): Amount of memory to be allocated in a given unit.
         - unit (str): The unit of memory.
     """
     try:
         matches = re.match(
             r"^(?P<value>[0-9]+(\.[0-9]+)?)(?P<unit>[^0-9])?$", max_memory
         )
-        value = float(matches["value"])
+        value = int(round(float(matches["value"])))
```
christophfink, line 82: round returns an int, casting is not necessary

christophfink: also, it would be nice to be able to supply decimal values: e.g., `1.5G`

src/r5py/util/memory_footprint.py
christophfink: If we go for integer output, I think the values should be bytes (and [`jvm.py`](https://github.com/r5py/r5py/blob/main/src/r5py/util/jvm.py#L28) adapted to accept bytes values (just the `M` removed), to then read 
```
        jpype.startJVM(
            f"-Xmx{MAX_JVM_MEMORY:d}",
            "-Xcheck:jni",
            classpath=[R5_CLASSPATH],
        )
```

src/r5py/util/memory_footprint.py
```diff
@@ -124,20 +124,20 @@ def _get_max_memory(max_memory):
...
         # convert to MiB
         if unit is None:
-            value *= 2**-20
+            value = value >> 20
         elif unit == "K":
-            value *= 2**-10
+            value = value >> 10
         elif unit == "M":
-            value *= 2**1
+            value = int(value * 1.024)
```
christophfink, line 131: What does this do? Is it converting from Megabytes to Mibibytes?

src/r5py/util/memory_footprint.py
```diff
@@ -49,14 +49,14 @@ def _share_of_ram(share=0.8, leave_at_least=(2 * (2**10))):
...
     Returns
     -------
-    float
+    int
         A value in MiB that is close to `share` portion of total RAM.
     """
     total_ram = psutil.virtual_memory().total / (2**20)
     if total_ram * (1.0 - share) > leave_at_least:
-        share_of_ram = round(share * total_ram)
+        share_of_ram = int(round(share * total_ram))
```
christophfink, line 57: `round()` returns `int`

src/r5py/util/memory_footprint.py
```diff
@@ -49,14 +49,14 @@ def _share_of_ram(share=0.8, leave_at_least=(2 * (2**10))):
...
     else:
-        share_of_ram = round(total_ram - leave_at_least)
+        share_of_ram = int(round(total_ram - leave_at_least))
```
christophfink, line 59: `round()` returns `int`

src/r5py/util/memory_footprint.py
HTenkanen: Hmm, you're right. The restriction to integer is indeed quite problematic. For example `"1000 MB"` would return 0 (found out now after a bit more testing) which is definitely not what we want. Will return the logic back to what it was. The problem why I started investigating this was due to the `"M"` conversion which actually returned incorrect output (it doubled the input).

src/r5py/util/memory_footprint.py
HTenkanen: Yes it converts Megabytes to Mibibytes. The original logic returned incorrect result.

src/r5py/util/memory_footprint.py
HTenkanen: True. Will convert the logic to what it was

src/r5py/util/memory_footprint.py
HTenkanen: Ha, true. `np.round()` then again returns float, which to me is actually more logical. Was surprised learning that the Python's own round() indeed work like this (i.e. returns integer).

HTenkanen:
@christophfink : Okay hopefully now the last time.. 🤞🏻 Could you please review this, and if okay, do a **Squash and merge**..was a messy one 🙈

src/r5py/util/memory_footprint.py
christophfink: It is inconsistent, in a way, because if called with a second argument (number of decimal places), it returns float ;)

christophfink:
I refactored the code a bit, also to have exceptions raised where they matter, and to have the helper functions be less specific (more generic). I also fixed the Mibibytes/Megabytes confusion, and converted all the (internal) values to bytes (so no unit suffix means no unit suffix in all places in the code).

There remain two edge cases for which it is very difficult to write tests for (they’re dynamic, and I cannot think of code that does not 1:1 reproduce the to-be-tested code), I added comments to that avail to the test file.

Tests now pass, coverage should go up :)

christophfink:
Oh I talked to quickly, we now have too little memory allocated on the github runners, apparently. I’ll check on Monday, the latest.

christophfink (commented):
I, myself, now changed a few things, still. @HTenkanen , could you take a loot at his, still? thanks!

HTenkanen:
@christophfink Nice, thanks! 👍🏻 Will check and come back to these things overall early next week. Now going for the weekend to Flow 🥳😎

HTenkanen (commented):
@christophfink : Checked this now, and looks good! So this is good to go on my part 👍🏻 Could you merge? 🙂

christophfink (approved):

PR #130, opened 2022-05-26
Make the JVM starting/Java imports more explicit, and call start_jvm() where it is needed, not everywhere

### LINKED ISSUE #94: Reduce reliance on module loading side effects
Currently, `import jvm` starts the JVM without calling anything else. This is not great, in a way, and we should maybe introduce a function `r5py.util.jvm.start_jvm()` that gets called in the relevant source files (before importing java classes; restriction connected to jpype’s architecture), and that ‘fails gracefully’ if the JVM has been started already. 

This would also help with the import order of relative imports (make the order less significant).

christophfink:
closes #94

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -3,3 +3,7 @@ max-line-length = 88
 select = C,E,F,W,B,B950
 ignore = E203, E501, W503
 exclude = docs/conf.py
+
+per-file-ignores = """
+    src/r5py/r5/__init__.py: E402
+"""
--- a/.flake8
+++ b/.flake8
@@ -3,7 +3,3 @@ max-line-length = 88
 select = C,E,F,W,B,B950
 ignore = E203, E501, W503
 exclude = docs/conf.py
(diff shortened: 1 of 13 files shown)


christophfink:
so because the last commit has ‘[skip ci]’ (I added it only after opening the pull request), the checks are skipped altogether?

christophfink:
The code passes tests locally, please have a quick review first @HTenkanen , then we might get another commit, anyway ;)

HTenkanen:
@christophfink Yes, can take a look later today. 👍🏻 Now heading for PhD defense of one of our students 🎩

christophfink:
Sorry for rewriting history, pushed to wrong branch

HTenkanen (approved):
Looks good! 👍🏻 Don't understand much about the details of Jpype stuff in `jvm.py` but everything works fine, so can't complain. 😄

PR #131, opened 2022-05-26
Fix incorrect conversion of bytes, refactor and add docstrings.

### LINKED ISSUE #129: `max_memory` without specified unit expects the input in kilobytes?
I believe these lines [expect the input in kilobytes](https://github.com/r5py/r5py/blob/main/src/r5py/util/memory_footprint.py#L72-L76)? Is this expected behavior? 

It would be good to improve the [docstring](https://github.com/r5py/r5py/blob/main/src/r5py/util/memory_footprint.py#L64) about expected/possible inputs.

HTenkanen:
Resolves #129

### DIFF
--- a/src/r5py/util/memory_footprint.py
+++ b/src/r5py/util/memory_footprint.py
@@ -73,9 +73,9 @@ def parse_max_memory_string(max_memory):
 
         if unit is None:
             raise ValueError(
-                f"Could not interpret the memory unit from --max-memory."
-                f"The suffix for --max-memory should be '%', 'K', 'M', 'G' or 'T'."
-                f"For example to allocate five gigabytes of memory, use: '5G'"
+                "Could not interpret the memory unit from --max-memory."
+                "The suffix for --max-memory should be '%', 'K', 'M', 'G' or 'T'."
+                "For example to allocate five gigabytes of memory, use: '5G'"
             )
         return value, unit
     except TypeError:
--- a/src/r5py/util/memory_footprint.py
+++ b/src/r5py/util/memory_footprint.py
@@ -76,7 +76,9 @@ def _parse_max_memory_string(max_memory):
         - unit (str): The unit of memory.
     """
     try:
-        matches = re.match(r"^(?P<value>[0-9]+(\.[0-9]+)?)(?P<unit>[^0-9])?$", max_memory)
+        matches = re.match(
+            r"^(?P<value>[0-9]+(\.[0-9]+)?)(?P<unit>[^0-9])?$", max_memory
+        )
         value = float(matches["value"])
         unit = matches["unit"]
 
--- a/src/r5py/util/memory_footprint.py
+++ b/src/r5py/util/memory_footprint.py
@@ -76,7 +76,7 @@ def _parse_max_memory_string(max_memory):
         - unit (str): The unit of memory.
     """
     try:
-        matches = re.match(r"(?P<value>[0-9]+(\.[0-9]+)?)(?P<unit>[^0-9])?", max_memory)
+        matches = re.match(r"^(?P<value>[0-9]+(\.[0-9]+)?)(?P<unit>[^0-9])?$", max_memory)
         value = float(matches["value"])
         unit = matches["unit"]
 
--- a/src/r5py/util/memory_footprint.py
+++ b/src/r5py/util/memory_footprint.py
@@ -25,14 +25,14 @@ config.argparser.add(
         Use % as a suffix to specify a share of total RAM;
         K, M, G, T to specify KiB, MiB, GiB, or TiB, respectively.
         Values are rounded to the closest MiB.
-        Values without suffix will raise an error.
+        Values without suffix are interpreted as bytes.
     """,
     default="80%",
 )
 arguments = config.arguments()
 
 
-def share_of_ram(share=0.8, leave_at_least=(2 * (2**10))):
+def _share_of_ram(share=0.8, leave_at_least=(2 * (2**10))):
     """
     Calculate a share of total RAM.
 
@@ -60,18 +60,16 @@ def share_of_ram(share=0.8, leave_at_least=(2 * (2**10))):
     return share_of_ram
 
 
-def parse_max_memory_string(max_memory):
+def _parse_max_memory_string(max_memory):
     """
     Extract maximum memory value and unit from text input.
     """
     try:
-        matches = re.match(
-            r"(?P<value>[0-9]+(\.[0-9]+)?)(?P<unit>[%KMGT])?", max_memory
-        )
+        matches = re.match(r"(?P<value>[0-9]+(\.[0-9]+)?)(?P<unit>[^0-9])?", max_memory)
         value = float(matches["value"])
         unit = matches["unit"]
 
-        if unit is None:
+        if unit is None and unit not in "%KMGT":
             raise ValueError(
                 "Could not interpret the memory unit from --max-memory."
                 "The suffix for --max-memory should be '%', 'K', 'M', 'G' or 'T'."
@@ -85,7 +83,7 @@ def parse_max_memory_string(max_memory):
         )
 
 
-def get_max_memory(max_memory):
+def _get_max_memory(max_memory):
     """
     Interpret the config parameter --max-memory.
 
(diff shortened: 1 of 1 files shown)


src/r5py/util/memory_footprint.py
```diff
@@ -60,29 +60,64 @@ def share_of_ram(share=0.8, leave_at_least=(2 * (2**10))):
...
+    value, unit = parse_max_memory_string(max_memory)
+
+    if unit == "%":
+        max_memory = share_of_ram(share=(value / 100.0))
+    else:
+        # convert to MiB
+        if unit == "K":
+            value *= 2**-20
+            if value < 1:
+                value = 1
```
christophfink, line 113: I think this logic (`value = max(value, 1)`) should be moved to after the `round()`ing.

src/r5py/util/memory_footprint.py
```diff
@@ -60,29 +60,64 @@ def share_of_ram(share=0.8, leave_at_least=(2 * (2**10))):
...
-                    value = 1
-            # elif unit == "M":
-            #    value *= 2 ** 1
-            elif unit == "G":
-                value *= 2**10
-            elif unit == "T":
-                value *= 2**20
-            max_memory = round(value)
+
+        if unit is None:
```
christophfink, line 74: I’m not really happy to kick out the possibility to specify this in bytes, see my (very easy to implement) suggestion in the relating issue. Use case: e.g. when this value is computed somewhere else (no need to complicatedly convert to +suffix, and back again)

HTenkanen: Yes, we can do this (see my comment in the issue). However, what is the use-case when someone would really request memory in bytes? 🙂 At least, it would probably make sense to use some other unit as the default, no? Like `MiB`

src/r5py/util/memory_footprint.py
HTenkanen: Done.

src/r5py/util/memory_footprint.py
```diff
@@ -60,29 +60,64 @@ def share_of_ram(share=0.8, leave_at_least=(2 * (2**10))):
...
+        return value, unit
     except TypeError:
-        raise ValueError(f"Could not interpret --max-memory: {max_memory}")
+        raise ValueError(
+            f"Could not interpret --max-memory: {max_memory}."
+            f"To allocate memory, use e.g. '5G' for five gigabytes of memory."
+        )
+
+
+def get_max_memory(max_memory):
```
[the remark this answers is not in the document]
HTenkanen: > What I’m thinking is whether we should make all of the functions private, in fact? (i.e. pseudo-private, prefixed with an underscore), and only leave the constant MAX_JVM_MEMORY a public member of this module?

Yes, we can do this, and would make sense because these shouldn't be needed by a normal user 👍🏻

src/r5py/util/memory_footprint.py
christophfink: See my comment in the issue, I would find it confusing is the unprefixed unit is anything else but the base unit (bytes). 

But I somehow question the (my) choice of parsing into MiB here. There’s no real reason for it, and in fact it would be weird if we accept bytes, but then pass on a value that’s not bytes (without renaming the constant). The JVM accepts bytes perfectly fine, so why not stay there, in a way. That would also allow to make the if/elif tree much more DRY (the suffix changes the exponent, only).

christophfink:
Can you maybe commit the test for `memory_footprint.py` into the same pull request?

src/r5py/util/memory_footprint.py
HTenkanen: > See my comment in the issue, I would find it confusing is the unprefixed unit is anything else but the base unit (bytes).
> 
> But I somehow question the (my) choice of parsing into MiB here. There’s no real reason for it, and in fact it would be weird if we accept bytes, but then pass on a value that’s not bytes (without renaming the constant). The JVM accepts bytes perfectly fine, so why not stay there, in a way. That would also allow to make the if/elif tree much more DRY (the suffix changes the exponent, only).

Okay, well let's make a separate issue about this change. Will now need to finish with this one.

src/r5py/util/memory_footprint.py
```diff
@@ -60,29 +60,64 @@ def share_of_ram(share=0.8, leave_at_least=(2 * (2**10))):
...
+    """
+
+    value, unit = parse_max_memory_string(max_memory)
+
+    if unit == "%":
+        max_memory = share_of_ram(share=(value / 100.0))
+    else:
+        # convert to MiB
+        if unit == "K":
+            value *= 2**-20
```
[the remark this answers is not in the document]
christophfink: In the comment we talked about bytes ;)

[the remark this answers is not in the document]
HTenkanen: I guess you mean `value *= 2**-10`?  

`2**10` wouldn't make sense because it's the one converting Gigabytes.

HTenkanen:
@christophfink : Okay, now changed things accordingly. Could you check?

src/r5py/util/memory_footprint.py
```diff
@@ -60,29 +60,64 @@ def share_of_ram(share=0.8, leave_at_least=(2 * (2**10))):
...
-                    value = 1
-            # elif unit == "M":
-            #    value *= 2 ** 1
-            elif unit == "G":
-                value *= 2**10
-            elif unit == "T":
-                value *= 2**20
-            max_memory = round(value)
+
+        if unit is None and unit not in "%KMGT":
```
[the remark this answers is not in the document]
HTenkanen: Oops..

src/r5py/util/memory_footprint.py
```diff
@@ -60,29 +60,64 @@ def share_of_ram(share=0.8, leave_at_least=(2 * (2**10))):
...
+        Values without suffix are interpreted as bytes.
+    """
+
+    value, unit = _parse_max_memory_string(max_memory)
+
+    if unit == "%":
+        max_memory = _share_of_ram(share=(value / 100.0))
+    else:
+        # convert to MiB
+        if unit == "K":
```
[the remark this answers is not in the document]
HTenkanen: ✅

src/r5py/util/memory_footprint.py
```diff
@@ -60,29 +60,64 @@ def share_of_ram(share=0.8, leave_at_least=(2 * (2**10))):
...
+    Arguments
+    ---------
+
+    max_memory : str
+        Memory limit for the JVM running R5.
+
+        Use % as a suffix to specify a share of total RAM;
+        K, M, G, T suffix specify KiB, MiB, GiB, or TiB, respectively.
+        Values are rounded to the closest MiB.
+        Values without suffix are interpreted as bytes.
```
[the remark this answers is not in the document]
HTenkanen: ✅

src/r5py/util/memory_footprint.py
```diff
@@ -60,29 +60,64 @@ def share_of_ram(share=0.8, leave_at_least=(2 * (2**10))):
     return share_of_ram
 
 
-def max_memory(max_memory):
-    """Interpret the config parameter --max-memory."""
+def _parse_max_memory_string(max_memory):
+    """
+    Extract maximum memory value and unit from text input.
+    """
```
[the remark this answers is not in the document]
HTenkanen: ✅

src/r5py/util/memory_footprint.py
```diff
@@ -60,29 +60,82 @@ def share_of_ram(share=0.8, leave_at_least=(2 * (2**10))):
...
+
+    Returns
+    -------
+    tuple: a tuple containing
+        - value (float): Amount of memory to be allocated in a given unit.
+        - unit (str): The unit of memory.
+    """
     try:
-        matches = re.match(r"(?P<value>[0-9]+(\.[0-9]+)?)(?P<unit>[%MGT])?", max_memory)
+        matches = re.match(r"(?P<value>[0-9]+(\.[0-9]+)?)(?P<unit>[^0-9])?", max_memory)
```
christophfink, line 79: The regex should only match to the entire string. So, it should read `r"^(?P<value>[0-9]+(\.[0-9]+)?)(?P<unit>[^0-9])?$"`

(Sorry, this is my bad - just tried it out the first time now that you submitted the PR)

christophfink: I’ll push the change myself, quickly, so you don’t have to wait another round.

PR #135, opened 2022-05-26
Do not fail entire test if codecov upload action fails

### LINKED ISSUE #134: Codecov upload fails entire test
In the lint-test-etc.yml we use codecov’s actions to upload our results to their page. This upload action sometimes fails (maybe a race condition of several of the matrix tests finishing concurrently). Let’s configure codecov in a way that this does not fail the test (it’s the last step, anyway).

christophfink:
closes #134

### DIFF
--- a/.github/workflows/lint-test-upon-push.yml
+++ b/.github/workflows/lint-test-upon-push.yml
@@ -86,5 +86,5 @@ jobs:
       - name: Update codecov.io
         uses: codecov/codecov-action@v3.1.0
         with:
-          fail_ci_if_error: true
+          fail_ci_if_error: false
           token: ${{ secrets.CODECOV_TOKEN }}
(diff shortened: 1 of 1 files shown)


christophfink:
Hm, this not seem to fix things. Seems like it’s an upstream bug: https://github.com/codecov/codecov-action/issues/745

christophfink:
upstream bug should be resolved, let’s give this a try

christophfink:
This was now fixed upstream, and it seems to work. 
I would still advocate to keep `fail_ci_if_error` at its default value `false`, as the updating of codecov is not exactly critical, and seems to suffer from hickups. What’s your take on this, @HTenkanen

HTenkanen:
> This was now fixed upstream, and it seems to work. I would still advocate to keep `fail_ci_if_error` at its default value `false`, as the updating of codecov is not exactly critical, and seems to suffer from hickups. What’s your take on this, @HTenkanen?

Yep, I have noticed that codecov is a bit unstable, so would also vote for keeping the default (i.e. `false`)

HTenkanen (approved):

Issue #140, opened 2022-06-01
Check that input data has unique id columns


wklumpen:
We should make sure the input origins for travel time computation includes unique ID columns. Just tried to run it with non-unique IDs and got the following error:

```
  File "...\pandas\core\series.py", line 191, in wrapper
    raise TypeError(f"cannot convert the series to {converter}")
TypeError: cannot convert the series to <class 'float'>
```
I wonder if we should provide an option for the user to specify the unique ID column also.

HTenkanen:
Yeah, we might want to add a couple of assertions to ensure the ids are unique. Also adding possibility to define the id column is a good idea to give a bit of flexibility to the user 👍🏻 

In addition, I'm thinking whether it would make sense to automatically create the ids based on row numbers. Perhaps not (can probably cause more issues/confusion than help), but it would be a rather trivial thing to do.

wklumpen:
> In addition, I'm thinking whether it would make sense to automatically create the ids based on row numbers. Perhaps not (can probably cause more issues/confusion than help), but it would be a rather trivial thing to do.

The danger here is we generate TT matrices with auto-generated IDs and the user thinks it's referring to something else? If we're going to do this I would suggest being pretty obvious about it like `auto01` or something

HTenkanen:
> > In addition, I'm thinking whether it would make sense to automatically create the ids based on row numbers. Perhaps not (can probably cause more issues/confusion than help), but it would be a rather trivial thing to do.
> 
> The danger here is we generate TT matrices with auto-generated IDs and the user thinks it's referring to something else? If we're going to do this I would suggest being pretty obvious about it like `auto01` or something

Perhaps we could skip this option for now at least. I see it probably causing more harm than benefits.

christophfink:
Reverted the merge, this should not have gone into main, yet

christophfink:
I removed it from the milestone, instead, if that was your motivation @HTenkanen ?

PR #141, opened 2022-06-01

wklumpen:
See #140 

I see also somehow that the hash assertion code has stuck around (trying to keep branches separate) but it looks like it's just the parentheses fix is there.


src/r5py/r5/travel_time_matrix_computer.py
```diff
@@ -45,6 +45,14 @@
 }
 
 
+class NoIDColumnError(Exception):
```
HTenkanen, line 48: Not sure but should we perhaps have a separate file for these errors? That's what I've typically done. Thoughts, @christophfink?

src/r5py/r5/travel_time_matrix_computer.py
```diff
@@ -92,10 +100,24 @@ def __init__(
             transport_network = TransportNetwork(*transport_network)
         self.transport_network = transport_network
 
+        # Quick validation to ensure an ID column exists
+        if "id" not in origins.columns:
```
HTenkanen, line 104: As discussed in the issue, I would perhaps also add the possibility to specify a column which is used as unique id. It's a bit silly restriction to always expect 'id' column to be found as it can be called anything.

src/r5py/util/classpath.py
```diff
@@ -36,7 +36,7 @@
     R5_CLASSPATH = str(pathlib.Path(CACHE_DIR) / pathlib.Path(R5_JAR_URL).name)
     try:
         with open(R5_CLASSPATH, "rb") as jar:
-            assert hashlib.sha256(jar.read()).hexdigest == R5_JAR_SHA256
+            assert hashlib.sha256(jar.read()).hexdigest() == R5_JAR_SHA256
```
HTenkanen, line 39: Shouldn't this be in a separate PR? Did it slip here?

HTenkanen: Ah, I see you commented about this already

HTenkanen:
@wklumpen Looks good, a few comments. Would you also write tests for this unique id issue? Let's try not to get further behind with the tests because we already now have quite a lot of catching up to do.

HTenkanen (changes requested):

wklumpen:
> @wklumpen Looks good, a few comments. Would you also write tests for this unique id issue? Let's try not to get further behind with the tests because we already now have quite a lot of catching up to do.

I can certainly do a test - the problem is loading (or creating) and invalid dataset with a test. I suppose I could generate a simple invalid GeoJSON on the fly?

Maybe a larger question going forward: Should tests be attached to the PRs or added as a separate task?

src/r5py/util/classpath.py
wklumpen: Yep. If I could make Git forget all about it I would....

src/r5py/r5/travel_time_matrix_computer.py
wklumpen: Happy to do that - suggestions on naming convention/location?

`util/errors.py` perhaps?

christophfink: sounds great! 

Can we: 
- call the file `util/exceptions.py` (I’d expect more to come), and 
- inherit from 
    - a more specific exception (maybe `ValueError`?), **and** 
    - a generic R5pyException (so we have the possibility to catch all r5py errors with `except R5pyException:` )

src/r5py/r5/travel_time_matrix_computer.py
christophfink: I guess a separate issue or PR would be a good place :+1:

src/r5py/r5/travel_time_matrix_computer.py
```diff
@@ -92,10 +100,24 @@ def __init__(
...
+            raise NoIDColumnError("Origin dataset must contain an 'id' column.")
+        # And to make sure the column is indeed unique
+        if len(origins.id.unique()) < origins.shape[0]:
```
christophfink, line 107: This would be more readable, and probably also better performing: https://pandas.pydata.org/docs/reference/api/pandas.Series.is_unique.html

src/r5py/r5/travel_time_matrix_computer.py
```diff
@@ -92,10 +100,24 @@ def __init__(
...
+            raise NotUniqueIDError("Origin id values must be unique.")
         self.origins = origins
 
         if destinations is None:
             destinations = origins
+
+        # Quick validation to ensure an ID column exists
```
christophfink, line 114: Could you take this entire block into a utility function that can be called from `origins()` and from `destinations()`, and possibly be reused in other classes in the future?

src/r5py/r5/travel_time_matrix_computer.py
HTenkanen: Yep, sounds good 👍🏻

HTenkanen:
> I can certainly do a test - the problem is loading (or creating) and invalid dataset with a test. I suppose I could generate a simple invalid GeoJSON on the fly?

Yes, you can create a simple dummy dataset to reproduce the issue (i.e. have non-unique values) and then use that to test that r5py handles it correctly.
 
> Maybe a larger question going forward: Should tests be attached to the PRs or added as a separate task?

Typically tests are attached to the same PR, and that's how we also suggest to do in our contribution guide, so let's follow this convention: https://r5py.readthedocs.io/en/latest/CONTRIBUTING.html#making-changes-and-writing-tests

src/r5py/r5/travel_time_matrix_computer.py
christophfink: This utility function could be a private method of whatever should be the parent class of all ‘worker classes’, but for the time being a separate file in `utils/` would probably do the job.

christophfink:
> Maybe a larger question going forward: Should tests be attached to the PRs or added as a separate task?
In my opinion, going forward, we should aim to have a test with every new addition of code. Adding a test to existing functionality, then, should be a separate issue+PR?

christophfink:
> I can certainly do a test - the problem is loading (or creating) and invalid dataset with a test. I suppose I could generate a simple invalid GeoJSON on the fly?

I guess just invalidating it in the test would be good? so dropping the id column, renaming the id column, making it non-unique. If that happens in the test it’s clear how the test file was modified, and does not tempt anybody to try to use the corrupt asset? (just my 2c)

wklumpen:
It wouldn't be difficult to generate a dataset in memory during the test in this case (3 rows of data let's say) - while that's not feasible for every test case, should we try and do that approach where possible instead of relying on specific files?

christophfink:
Well, @HTenkanen definitely has a point, when he says that it would be good to have static test datasets that do not change. After all, the same error might affect test data generation and the code that is tested. 

Should we start a new directory with test data in test/data/ or so?

wklumpen:
Not sure why it's failing the tests - trying to import `geopandas` seems to fail, but we definitely require geopandas?

HTenkanen:
> Not sure why it's failing the tests - trying to import `geopandas` seems to fail, but we definitely require geopandas?

@wklumpen : We are not actually requiring geopandas at the moment. It was part of the dev-environment mainly for the purpose of documentation. There was some discussion about the required packages in #52. 

But for real, I think it is only a matter of time when we will have something developed in the source code which will require geopandas, so I'd say it would be okay to add `geopandas` as a dependency for all the environments in `ci/` directory: https://github.com/r5py/r5py/tree/main/ci

src/r5py/r5/travel_time_matrix_computer.py
christophfink: Thinking about it, I would argue against a configurable column name. It’s straight-forward for the user to rename a column, at the same time the code to handle dynamic column names necessarily will be fairly bloated. With good documentation, requiring specific column names is a very pythonic way of handling this.

christophfink:
> > Not sure why it's failing the tests - trying to import `geopandas` seems to fail, but we definitely require geopandas?
> 
> @wklumpen : We are not actually requiring geopandas at the moment. It is part of the dev-environment mainly for the purpose of documentation. There was some discussion about the required packages in #52.
> 
> But for real, I think it is only a matter of time when we will have something developed in the source code which will require geopandas (and now we already do need it for tests), so I'd say it would be okay to add `geopandas` as a dependency for all the environments in `ci/` directory: https://github.com/r5py/r5py/tree/main/ci
> 
> Ping @christophfink

Absolutely, `geopandas` is basically necessary to run anything (e.g., some inputs to the `TravelTimeMatrixComputer` are supposed to be `geopandas.GeoDataFrame`s), let’s just add it to the general dependencies. I’ll open a separate issue+PR for this.

wklumpen:
I may have lost the trail on this one - are we just waiting on the weird Windows issues for this PR?

christophfink:
> I may have lost the trail on this one - are we just waiting on the weird Windows issues for this PR?

weirdly, it seems that we would have to, before merging. Does not really mean you cannot push any other changes in the meantime :)

wklumpen:
"I can't do any work I'm on Windows" is the new "my code's compiling"

src/r5py/r5/travel_time_matrix_computer.py
HTenkanen: I guess this comment is resolved from this PR. Whether or not to allow configurable column name can be discussed in a separate issue / PR.

src/r5py/r5/travel_time_matrix_computer.py
HTenkanen: @christophfink @wklumpen: Let's turn this into a separate issue, so that we can get this PR rolling.

src/r5py/r5/travel_time_matrix_computer.py
HTenkanen: Added a separate issue for this in #162

HTenkanen:
@wklumpen and @christophfink, I went through the comments and created a couple of new issues based on some of them that required action, so we can resolve them later.

HTenkanen (approved):

christophfink:
@HTenkanen this was really not ready to merge! It’s fine to split it into new issues, but this branch did not work as expected, so should not go into main!

HTenkanen:
@christophfink Hmm, okay, it passed all the tests and didn't spot anything problematic, so what is the issue with the PR?

christophfink:
All the unresolved comments went straight into main!

christophfink:
Removing them from this PR would also mean removing the code that added them.

PR #148, opened 2022-08-15
Add `geopandas` as a requirement

christophfink:
(while I was at it, I also moved the test requirements from the deprecated `setup.cfg` entry `test_requires` to `pyproject.toml`)


HTenkanen (approved):
Looks good! 💪🏻

Issue #147, opened 2022-08-15
Add `geopandas` as a requirement/dependency


christophfink:
see below, but tl;dr: technically, r5py does not require geopandas, however, anybody using it will need it, and tests will be easier to write knowing it’s around, to.

> > Not sure why it's failing the tests - trying to import `geopandas` seems to fail, but we definitely require geopandas?
> 
> @wklumpen : We are not actually requiring geopandas at the moment. It is part of the dev-environment mainly for the purpose of documentation. There was some discussion about the required packages in #52.
> 
> But for real, I think it is only a matter of time when we will have something developed in the source code which will require geopandas (and now we already do need it for tests), so I'd say it would be okay to add `geopandas` as a dependency for all the environments in `ci/` directory: https://github.com/r5py/r5py/tree/main/ci
> 
> Ping @christophfink

Absolutely, `geopandas` is basically necessary to run anything (e.g., some inputs to the `TravelTimeMatrixComputer` are supposed to be `geopandas.GeoDataFrame`s), let’s just add it to the general dependencies. I’ll open a separate issue+PR for this.

_Originally posted by @christophfink in https://github.com/r5py/r5py/issues/141#issuecomment-1214869161_

christophfink:
So, apparently, on the GitHub Windows runners, geopandas explicitely needs `fiona` or `pyogrio` installed. Seems we need to add that to the requirements/environments.

PR #151, opened 2022-08-16
add fiona to environments and requirements

christophfink:
see reopened issue #147

### DIFF
--- a/ci/python_310.yaml
+++ b/ci/python_310.yaml
@@ -6,6 +6,7 @@ channels:
 dependencies:
     - python=3.10
     - configargparse
+    - fiona
     - geopandas
     - joblib
     - jpype1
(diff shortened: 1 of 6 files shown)


jorisvandenbossche:
Getting here from @HTenkanen's comment in the geopandas gitter. But so in general this should not be needed that you have to add Fiona explicitly as dependency as well (and geopandas might actually drop fiona as a required dependency in the near future, which is also a reason to not do this here)

christophfink:
> Getting here from @HTenkanen's comment in the geopandas gitter. But so in general this should not be needed that you have to add Fiona explicitly as dependency as well (and geopandas might actually drop fiona as a required dependency in the near future, which is also a reason to not do this here)

I figured, but then sometimes conda seems to resolve dependencies in a weird way, and seemed like a likely cause. Turns out it it’s likely connected to pytest (but I see you commented in the discussion over in the pull request, so you’ve seen our next view steps searching in the dark). Thanks for checking in!

PR #163, opened 2022-09-07
Do not allow lower allocated memory than absolute minimum RAM in `share_of_ram()` function

### LINKED ISSUE #127: `share_of_ram()` possibly returns negative values
While writing the tests, spotted an issue. If user requests with `leave_at_least` parameter more memory than actually exists on the user's computer, the function will return a negative value:
https://github.com/r5py/r5py/blob/main/src/r5py/util/memory_footprint.py#L59

This probably doesn't cause any issues at the moment and this function is not something that users are likely to use for anything, but still I think fixing this logic would make sense: i.e. return 0 and perhaps also raise an exception if memory cannot be allocated to the tool at all.

HTenkanen:
Closes #127

### DIFF
--- a/src/r5py/util/memory_footprint.py
+++ b/src/r5py/util/memory_footprint.py
@@ -50,12 +50,22 @@ def _share_of_ram(share=0.8, leave_at_least=(2 * 1024**3)):
     -------
     int
         A value in bytes that is close to `share` portion of total RAM.
+        Note: Will return the absolute minimum RAM (200 MB) if lower value is requested.
     """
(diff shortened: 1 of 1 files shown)


christophfink:
But that’s already taken care of in https://github.com/r5py/r5py/blob/main/src/r5py/util/memory_footprint.py#L168

christophfink (commented):
This is already implemented, https://github.com/r5py/r5py/blob/main/src/r5py/util/memory_footprint.py#L168

PR #165, opened 2022-09-07
Check for unique ID columns in origins/destinations

christophfink:
See all comments in closed PR #141

PR #166, opened 2022-09-07

christophfink:
Continuing here with what was begun in PR #140 and #165


christophfink:
Waiting to merge #167 first, as its tests also use test data sets (and the paths should be taken into `conftest.py`)

HTenkanen (approved):
Looks good to me!

Issue #162, opened 2022-09-07
Use `is_unique()` method to check unique values


HTenkanen:
This would be more readable, and probably also better performing: https://pandas.pydata.org/docs/reference/api/pandas.Series.is_unique.html

_Originally posted by @christophfink in https://github.com/r5py/r5py/pull/141#discussion_r887804355_

Issue #161, opened 2022-09-07
Make a utility function to test unique-id


HTenkanen:
Could you take this entire block into a utility function that can be called from `origins()` and from `destinations()`, and possibly be reused in other classes in the future?

_Originally posted by @christophfink in https://github.com/r5py/r5py/pull/141#discussion_r887805874_

HTenkanen:
This utility function could be a private method of whatever should be the parent class of all ‘worker classes’, but for the time being a separate file in utils/ would probably do the job. 

https://github.com/r5py/r5py/pull/141/files#r887819309

PR #167, opened 2022-09-07

christophfink:
Continuing here with what was begun in PR #140 and #165


christophfink:
closes #161 
closes #162 
closes #140

src/r5py/util/check_od_data_set.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python3
```
HTenkanen, line 1: Might we also have other "checks" for files/inputs etc? Just thinking that should we have a more generic filename into which we could possibly add also other checks in the future. The file could e.g. be `validate_input.py` or something along those lines..

HTenkanen (approved):
One comment regarding the name of the file for you to consider, but that is not really critical.

src/r5py/util/check_od_data_set.py
christophfink: that’s true :+1:

christophfink: I went even bigger and called the file `data_validation.py`

PR #171, opened 2022-09-07
Fully support Python 3.8 (avoid newer language features)

### LINKED ISSUE #137: `TravelTimeMatrixComputer` uses features of Python>=3.9
The `|` operator for `dict`s, as used, e.g., in https://github.com/r5py/r5py/blob/7d5d4e4e32f1cf4114a6d6ea3c85fe9e3e513589/src/r5py/r5/travel_time_matrix_computer.py#L238, was only introduced in Python 3.9, but we aim to support >=3.8

christophfink:
closes #137

### DIFF
--- a/src/r5py/r5/travel_time_matrix_computer.py
+++ b/src/r5py/r5/travel_time_matrix_computer.py
@@ -227,21 +227,21 @@ class TravelTimeMatrixComputer:
         # return travel times only
 
         # create the columns in order to force dtypes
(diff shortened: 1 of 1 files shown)


christophfink:
This also got much more logical: _first_ add the `from_id` and `to_id` columns, only then the other columns (that need conditional naming)

HTenkanen (approved):
Yep, this is definitely easier to follow. :)

PR #172, opened 2022-09-07
Change installation instructions to not have users install openjdk from mamba

### LINKED ISSUE #170: Installation instructions: don’t install OpenJDK from mamba
In the light of issue #156 , let’s update the installation instructions to require users to install an OpenJDK manually, rather than using mamba.

christophfink:
closes #170

### DIFF
--- a/ci/python_310.yaml
+++ b/ci/python_310.yaml
@@ -6,7 +6,6 @@ channels:
 dependencies:
     - python=3.10
     - configargparse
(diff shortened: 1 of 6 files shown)


christophfink:
While I was at it, I also removed the explicit dependency on `fiona` from the environment files.

HTenkanen (approved):
Super! You're so fast! 😆

PR #179, opened 2022-09-08
Improve the basic tutorial

### LINKED ISSUE #80: Improve basic tutorial
(This is on a sidenote, and deserves its own issue, maybe)

We should also try to streamline the story line of the basic-usage notebook, I think first having the grid points, then loading the transport network, might be a better idea.

_Originally posted by @christophfink in https://github.com/r5py/r5py/issues/64#issuecomment-1125801130_

In addition, to this ^, it would be very nice to also show something visual from the results. I.e. create a map showing the travel times to/from specific point.

HTenkanen:
Closes #80 

I now added quite a bit of information about different ways to use `r5py`. I think it should now cover the basics fairly okay.


HTenkanen:
@christophfink : Could you please check? 🙂

HTenkanen:
Blah, some old commits seem to be visible here again.. Oh well.

HTenkanen:
Will make a clean PR without the old commits.

PR #180, opened 2022-09-09
Improve basic tutorial

### LINKED ISSUE #80: Improve basic tutorial

HTenkanen:
Closes #80 

I now added quite a bit of information about different ways to use `r5py`. I think it should now cover the basics fairly okay.

Could you @christophfink check?

### DIFF
--- a/docs/md/basic-usage.md
+++ b/docs/md/basic-usage.md
@@ -14,39 +14,131 @@ jupyter:
 
 # Usage
 
-```python nbsphinx="hidden" tags=["remove-cell"]
-# this cell is hidden from output
-# it’s used to 
-# - set sys.path to point to the local repo
-# - make ipython show all outputs of each cell
+## Introduction
 
-# - set sys.path to point to the local repo
-import pathlib
+**R5py** is a Python library for routing and calculating travel time matrices on multimodal transport networks (walk, bike, public transport and car).
+It provides a simple and friendly interface to R<sup>5</sup> (*the Rapid Realistic Routing on Real-world and Reimagined networks*) which is a [routing engine](https://github.com/conveyal/r5) developed by [Conveyal](https://conveyal.com/). `R5py` is designed to interact with [GeoPandas](https://geopandas.org) GeoDataFrames, and it is inspired by [r5r](https://ipeagit.github.io/r5r) which is a similar wrapper developed for R. `R5py` exposes some of R5’s functionality via its [Python API](reference.html), in a syntax similar to r5r’s. At the time of this writing, only the computation of travel time matrices has been fully implemented. Over time, `r5py` will be expanded to incorporate other functionalities from R5. 
+
+<!-- #region -->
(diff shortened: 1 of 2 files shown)


christophfink:
The test seems to fail in certain constellations:

```
tests/test_travel_time_matrix_computer.py:173: AssertionError
```

(see https://github.com/r5py/r5py/runs/8263907932?check_suite_focus=true#step:9:40 )

In particular, that means the following assertion fails:

```
assert travel_time_matrix["travel_time"].max() in [50, 51]
```

which of course is a bit worrying, since it means that even with not-frequency-based input data, R5 is not entirely deterministic.

You already have a range of values, could you check whether a broader range would help, or whether it’s entirely random?

HTenkanen:
> The test seems to fail in certain constellations:
> 
> ```
> tests/test_travel_time_matrix_computer.py:173: AssertionError
> ```
> 
> (see https://github.com/r5py/r5py/runs/8263907932?check_suite_focus=true#step:9:40 )
> 
> In particular, that means the following assertion fails:
> 
> ```
> assert travel_time_matrix["travel_time"].max() in [50, 51]
> ```
> 
> which of course is a bit worrying, since it means that even with not-frequency-based input data, R5 is not entirely deterministic.
> 
> You already have a range of values, could you check whether a broader range would help, or whether it’s entirely random?
> 
> Edit: the logs show the returned maximum value is 49, so it’s an off-by-one. I guess extending the range to include that could be enough? Still, thinking about future application, it is annoying that this is not deterministic.

Yeah, I have noticed this behavior, and it is indeed quite annoying that there seems to be fluctuation in the results. I will add +-3 minutes to the maximum value. Hopefully that would be enough, so that this wouldn't be a pain in the *** in the future.

HTenkanen:
@christophfink FYI: Here is a build for the docs on my fork: https://r5py-fork.readthedocs.io/en/fix_docs_build/notebooks/basic-usage.html 

It seems, that installing `matplotlib` is causing issues at the moment in the RTD environment, so we probably need to remove that dependency for the time being from the `requirements.txt`, so that we can get the docs up..

HTenkanen:
@christophfink : Okay, now the tests seem to pass again. Needed to add +-4 minutes to the maximum value. This fluctuation comes from all-to-all analyses, so there are 92x92 connections analyzed with the 1 hour time-window, so I guess we can expect a bit of stochasticity in the results.

christophfink (approved):
This is great! Thanks for adding so much more detail.

PR #185, opened 2022-09-10
keep R5 from saving temporary files (or at least hide them from the user)

### LINKED ISSUE #44: keep R5 from saving temporary files into input directories
Figure out how to make R5 save the .mapdb elsewhere, and not next
to the input files (they could well be on a r/o filesystem, or in a dedicate
data folder that should not be littered with random cache files).

There are (at least) the following three options of doing that:
  1) Use a java agent/instrumentation api to monkey-patch
     com.conveyal.r5.transit.TransportNetwork.fromFiles where the path to the 
     .mapdb file is hardcoded. Not sure how easy that is, and whether jpype
     (properly) supports it. 
  2) Have a patched com.conveyal.r5.transit.TransportNetwork (or com.conveyal.osmlib.OSM)
     in the class path _before_ the r5-all.jar. Easy, but messy, potential maintenance hell.
  3) Copy the input file(s) to a temporary directory before reading them.
     By far the cleanest option, but have to check /tmp disk space (if it’s a ramdisk - the 
     default on most current linux installations - we cut into memory; might have to check how 
     long R5 wants to be able to access input files)

christophfink:
closes #44


src/r5py/r5/transport_network.py
```diff
@@ -41,15 +43,21 @@ def __init__(self, osm_pbf, gtfs=[], build_config={}):
         build_json : dict
             options accepted by TNBuilderConfig (including SpeedConfig)
         """
-        # TODO: Add TNBuilderConfig and SpeedConfig options to docstring
```
christophfink, line 44: I removed this TODO comment because it becomes obsolete with the next R5 version

src/r5py/r5/transport_network.py
```diff
@@ -110,6 +122,43 @@ def __exit__(self, exception_type, exception_value, traceback):
         """Exit context."""
         return False
 
+    @functools.cached_property
+    def _cache_directory(self):
+        cache_dir = (
+            pathlib.Path(config.CACHE_DIR) / f"{self.__class__.__name__:s}_{hash(self):x}"
```
christophfink, line 128: this creates directory names looking like `CACHE_DIR/TransportNetwork_1a23b4c5/` (the hash part *should* be unique during one run, as it is derived from, among other things, the memory address of the TransportNetwork instance)

tests/test_transport_network.py
```diff
@@ -76,3 +79,27 @@ def test_street_layer_java_object(self, transport_network):
     def test_timezone(self, transport_network, gtfs_timezone_helsinki):
         assert isinstance(transport_network.timezone, java.time.ZoneId)
         assert transport_network.timezone.toString() == gtfs_timezone_helsinki
+
+    def test_cache_directory(self, transport_network_files_tuple):
```
christophfink, line 83: Cannot use the already generated `transport_network_from_files` fixture because that seems to have some persistent links: garbage collection does not kick in, `__del__()` is not called within the test scope.

src/r5py/r5/transport_network.py
```diff
@@ -110,6 +122,44 @@ def __exit__(self, exception_type, exception_value, traceback):
         """Exit context."""
         return False
 
+    @functools.cached_property
+    def _cache_directory(self):
+        cache_dir = (
+            pathlib.Path(config.CACHE_DIR)
+            / f"{self.__class__.__name__:s}_{hash(self):x}"
```
christophfink, line 129: > this creates directory names looking like CACHE_DIR/TransportNetwork_1a23b4c5/ (the hash part should be unique during one run, as it is derived from, among other things, the memory address of the TransportNetwork instance)

src/r5py/r5/transport_network.py
HTenkanen: 👍🏻

src/r5py/r5/transport_network.py
```diff
@@ -110,6 +122,44 @@ def __exit__(self, exception_type, exception_value, traceback):
...
+        Returns
+        -------
+        pathlib.Path
+            The path to the copy or link created
+        """
+        # try to first create a symbolic link, if that fails (e.g., on Windows),
+        # copy the file to a cache directory
+        input_file = pathlib.Path(input_file)
+        destination_file = pathlib.Path(self._cache_directory / input_file.name)
+        try:
```
HTenkanen, line 157: Hmm, I'm not a big fan of implementing logic using try and catch. Could we write this in a way that we detect whether the operating system is Windows and then do the copy/paste and otherwise create a symbolic link?

tests/test_transport_network.py
HTenkanen: 👍🏻

tests/test_transport_network.py
```diff
@@ -76,3 +79,27 @@ def test_street_layer_java_object(self, transport_network):
...
+        transport_network = r5py.TransportNetwork(*transport_network_files_tuple)
+        cache_dir = transport_network._cache_directory
+        assert cache_dir.is_dir()
+        assert len(list(cache_dir.glob("*"))) > 0  # files have been copied/linked to cache
+        del transport_network  # not sure whether this works as expected
```
HTenkanen, line 88: The tests pass, so this seems to work? So could maybe remove the comment?

HTenkanen (approved):
A couple of things for you to consider. The copy/paste of files vs symlink logic is something that I think would be good to think further, as try/catch for implementing a logic can be a bit risky business. 

I now approve this, so that you don't need to wait for me with the revision, as I'm still on a worktrip and can be a bit slow in checking stuff.

src/r5py/r5/transport_network.py
christophfink: That’s now a philosophical one, and I disagree with your stance here: Symbolic links are the way to go, except if an error occurs because operating system conditions are not met. To catch this anticipated error using an exception and in the exception attempting the (more fail-safe) second-best option is exactly the correct pattern, IMO.

This is at the core of the Pythonic philosophy of ‘Ask forgiveness, not permission’. Explicit logic for testing OS capabilities is not.

christophfink: Sorry to continue on this, but this is definitely the OPPOSITE of risky. This is trying the best-suited approach, and having a fallback if it is not possible.

tests/test_transport_network.py
christophfink: true, the last comment is superfluous, removing it

christophfink: In fact, I added a comment to the next line to explain what the two lines combined try to test

src/r5py/r5/transport_network.py
HTenkanen: Okay, fair enough 🙂 👍🏻

Issue #156, opened 2022-08-17
Fiona and OpenJDK don’t get along in Windows runners 


christophfink:
See the entire discussion in pull request https://github.com/r5py/r5py/pull/149.

It seems that Fiona/GDAL/OGR and OpenJDK somehow confuse each other on the GitHub Windows runners.

Could the people with access to ‘bare-metal’ Windows (@HTenkanen , @wklumpen ) please try to following minimum example:

1. Create an environment file `fionatest.yaml` with the following content:

```
name: fionatest

channels:
    - conda-forge

dependencies:
    - python=3.10
    - fiona
```

2. Use mamba to create a new environment using this file: 

```
mamba env create --file fionatest.yaml
```

3. try to import fiona (should be successful)

```
python -c "import fiona"
```

4. install openjdk (in this example using mamba, in the github runners we use the appropriate github action)

```
mamba install -y openjdk=11
```

5. try to import fiona, again (fails!)

```
python -c "import fiona"
```

6. Bonus step 1: uninstall the JDK

```
mamba remove -y openjdk
```

7. Bonus step 2: try to import fiona, again (STILL fails!)

```
python -c "import fiona"
```

Not quite sure how to get out of this, as we need *both* a JDK and fiona. At least I’m my semi-certain that it’s not pytest. I’m fairly sure that it’s somehow connected to environment variables (like %PATH%, e.g.), and that installing the JDK messes them up, apparently both with the GH action and from mamba. If you try the (hopefully reproducible) example, could you compar the environment variables (or at least %PATH%) before and after every step? (e.g., `python -c "import os; print(os.environ['PATH'])"`) Thanks!

_Originally posted by @christophfink in https://github.com/r5py/r5py/issues/149#issuecomment-1218061269_

HTenkanen:
@christophfink : Will try this out 👍🏻

HTenkanen:
Also as Joris hints here https://github.com/r5py/r5py/pull/149#issuecomment-1218091820,
a difference between environments is that **before** openjdk is installed, there actually isn't an environment variable called `JAVA_HOME` but openjdk adds that variable once it is installed. 👀

christophfink:
Does JAVA_HOME persist after uninstalling the JDK (it does not for me)

HTenkanen:
> Does JAVA_HOME persist after uninstalling the JDK (it does not for me)

Yes it does at least on my Windows.

HTenkanen:
[...]

PR #189, opened 2022-09-11
fix tests on windows

christophfink:
I might have figured out a way to fix issue #156. This works locally on my Windows VM, let’s see how far it goes on the GitHub runners. 

What I did was add an explicit import statement for `fiona` before `geopandas` in `conftest.py`. This does not yet fix the packaging issue with openjdk on conda-forge, but this is being dealt with upstream.

*EDIT:* The way I see it in retrospect, the issue was that geopandas attempted to import fiona from within a test case, which is isolated by pytest for good reasons. By importing fiona ahead of time, we avoid this problem.

### DIFF
--- a/ci/python_310.yaml
+++ b/ci/python_310.yaml
@@ -6,6 +6,7 @@ channels:
 dependencies:
     - python=3.10
     - configargparse
+    - fiona
     - geopandas
     - joblib
     - jpype1
--- a/.github/workflows/lint-test-upon-push.yml
+++ b/.github/workflows/lint-test-upon-push.yml
@@ -70,13 +70,6 @@ jobs:
           miniforge-version: latest
           use-mamba: true
 
-      - name: On Windows, resort JDK/GDAL conflicts
-        if: matrix.os == 'windows-latest'
-        run: mamba install --force-reinstall gdal libgdal libzip freetype
-
-      - name: Test whether fiona can be imported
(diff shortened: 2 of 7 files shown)


christophfink:
I also took the liberty of creating a new issue to reference the upstream bug in OpenJDK’s packaging for conda-forge, so we can close this biggie here but at the same time not loose track of the hopefully soon resolved upstream issue. 

#190

ci/python_310.yaml
```diff
@@ -6,6 +6,7 @@ channels:
 dependencies:
     - python=3.10
     - configargparse
+    - fiona
```
HTenkanen, line 9: `fiona` comes with geopandas always, so not necessarily needed here. However, I'm okay having it listed.

HTenkanen (approved):
Looks good! One comment regarding `fiona` for you to consider, but looks good overall!

ci/python_310.yaml
christophfink: That’s true. Let’s kick it from here, already `geopandas` itself is a bit of a stretch, since we do not (yet) import it anywhere (but it really IS needed for any sensible use of r5py)

christophfink: no, actually, thinking about it: we ARE now importing it in the tests, so I’d rather keep it included explicitely, so we don’t run into any issues with the test runners.

PR #194, opened 2022-09-16
Fix failing test on Windows, when temp files cannot be deleted

### LINKED ISSUE #193: `TransportNetwork.__del__()` test fails on Windows
```
 E           PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\runneradmin\\AppData\\Local\\r5py\\TransportNetwork_1ff8f07f6a\\kantakaupunki.osm.pbf.mapdb'
```

https://github.com/r5py/r5py/actions/runs/3066469908/jobs/4951701221

It seems that on Windows, something (the Java instance of the transport network?) still has the input file open when we try to destroy a `TransportNetwork`, and clean-up fails (we cannot delete the temp-file)

christophfink:
closes #193 

On Windows, the `TransportNetwork`’s cache files are still opened when we try to delete cache files ~in the instance destructor~.

### DIFF
--- a/ci/python_310.yaml
+++ b/ci/python_310.yaml
@@ -14,5 +14,3 @@ dependencies:
     - pandas
     - psutil
     - requests
-    # for debugging issue #193
-    - psutil
--- a/ci/python_310.yaml
+++ b/ci/python_310.yaml
@@ -14,3 +14,5 @@ dependencies:
     - pandas
     - psutil
     - requests
+    # for debugging issue #193
+    - psutil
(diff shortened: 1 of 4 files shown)


wklumpen (commented):
Is there a real reason to have this at all if `__del__()` takes care of them anyway? Or does it just make things smoother for other OS situations.

christophfink:
The reason that we don't use up too much space in the cache folder that might be a) on a ram disk, b) inside a (roaming) user profile, or c) on the main disk. a and b could potentially interfere with r5py/R5's performance directly, b) might cause transfer from/to file servers.

but yes, you definitely have a point there, as the early clean-up adds complexity to the code base and is not strictly necessary

PR #196, opened 2022-09-19
A few more tests

christophfink:
I added a few tests for ValidatingRequestsSession, and sorted the fixtures in conftest.py in alphabetical order


wklumpen (commented):
Looks good to me - I always wonder what to pick for invalid data - would it be a 'better' test if you took the valid SHA256 and just changed a few characters around? Mostly a curious question.

christophfink:
I have no real opinion on that TBH. I guess, the expected value in tests should preferably be hard-coded, but whether the invalid comparison value is far off or close to the valid value doesn't really make a difference? Not sure whether there's some convention around it?

PR #205, opened 2022-09-28
Fix routing for cars

### LINKED ISSUE #191: Calculations with CAR do not work with example data
@christophfink : Following the examples and using our sample data does not seem to work when using `CAR` as the travel mode (I assume this should work?):

```python
import geopandas 
import datetime
from r5py import TravelTimeMatrixComputer, TransitMode, LegMode, TransportNetwork

points_url = "https://github.com/r5py/r5py/raw/main/docs/data/Helsinki/population_points_2020.gpkg"
points = geopandas.read_file(points_url)
origin = points.loc[points["id"] == 54].copy()

transport_network = TransportNetwork(
    "../data/Helsinki/kantakaupunki.osm.pbf",
    [
        "../data/Helsinki/GTFS.zip"
    ]
)

travel_time_matrix_computer = TravelTimeMatrixComputer(
    transport_network,
    origins=origin,
    destinations=points,
    departure=datetime.datetime(2022,2,22,8,30),
    transport_modes=[LegMode.CAR]
)
travel_time_matrix = travel_time_matrix_computer.compute_travel_times()
travel_time_matrix.head()
```

This produces following error (partially truncated):
image

image

christophfink:
Found the culprit, why routing for cars kept failing: the default speed for roads that have not set any road class and/or speed limit is not copied from the Java `SpeedConfig`. Setting it on the Python side makes things work as expected :)

closes #191

### DIFF
--- a/src/r5py/r5/speed_config.py
+++ b/src/r5py/r5/speed_config.py
@@ -33,6 +33,7 @@ class SpeedConfig(dict):
             "tertiary_link": 25,
             "living_street": 5,
         },
+        "defaultSpeed": 40
     }
 
     def __init__(self, **kwargs):
--- a/src/r5py/r5/speed_config.py
+++ b/src/r5py/r5/speed_config.py
@@ -33,7 +33,7 @@ class SpeedConfig(dict):
             "tertiary_link": 25,
             "living_street": 5,
         },
-        "defaultSpeed": 40
+        "defaultSpeed": 40,
     }
 
     def __init__(self, **kwargs):
--- a/src/r5py/r5/speed_config.py
+++ b/src/r5py/r5/speed_config.py
@@ -33,7 +33,7 @@ class SpeedConfig(dict):
             "tertiary_link": 25,
             "living_street": 5,
         },
-        "defaultSpeed": 40,
+        "defaultSpeed": 25,
     }
 
     def __init__(self, **kwargs):


HTenkanen (commented):
Looks good! However, I now see that we don't seem to have tests for `speed_config.py`? I think adding those would be a good thing.

HTenkanen:
Also adding a test for routing with car, bike and walking would be good as we are currently only testing transit. We could perhaps make a separate issue / PR for the bike and walking tests if seems more logical / clean.

christophfink:
> Also adding a test for routing with car, bike and walking would be good as we are currently only testing transit. We could perhaps make a separate issue / PR for the bike and walking tests if seems more logical / clean.

Yes, that would be good, but probably something to move into a separate issue. I’ll open one

HTenkanen:
> Looks good! However, I now see that we don't seem to have tests for `speed_config.py`? I think adding those would be a good thing.

@christophfink Did you plan to work on these speed_config tests on this PR, or how is it? Just that I know what is the plan 🙂

christophfink:
> > Looks good! However, I now see that we don't seem to have tests for `speed_config.py`? I think adding those would be a good thing.
> 
> @christophfink Did you plan to work on these speed_config tests on this PR, or how is it? Just that I know what is the plan slightly_smiling_face

No, because `SpeedConfig` is removed in the next R5 version, not really worth adding tests on our side for it.

HTenkanen:
> > > Looks good! However, I now see that we don't seem to have tests for `speed_config.py`? I think adding those would be a good thing.
> > 
> > 
> > @christophfink Did you plan to work on these speed_config tests on this PR, or how is it? Just that I know what is the plan slightly_smiling_face
> 
> No, because `SpeedConfig` is removed in the next R5 version, not really worth adding tests on our side for it.

Ah right, so I guess then this is ready.

HTenkanen (approved):

PR #209, opened 2022-09-28
test different jdk origins

christophfink:
Run unit tests also separately depending on which origin OpenJDK was installed from (using mamba, or system-wide)


.github/workflows/lint-test-upon-push.yml
```diff
@@ -71,12 +75,14 @@ jobs:
...
+      - name: Install JVM (using mamba)
+        if: ${{ matrix.jdk_from == 'mamba' }}
+        run: mamba install 'openjdk>=11,<12'
+
       - name: Make sure pytest{,-cov,-lazy-fixture} are installed
-        shell: bash -l {0}
         run: mamba install -c conda-forge pytest pytest-cov pytest-lazy-fixture
 
       - name: Check environment
-        shell: bash -l {0}
```
christophfink, line 79: (note) shell is set to `bash -l {0}` by default in line 39

christophfink:
So in theory, this works, but there has been a regressing upstream with respect to how openjdk11 is packaged for conda-forge: https://github.com/conda-forge/openjdk-feedstock/issues/107

christophfink:
upstream has release a fix, let’s try this again

christophfink:
This should be fixed in #208 , let’s first merge that into main, and then main into here.

PR #217, opened 2022-10-04
A few of the tests that are still missing

christophfink:


christophfink:
On top of providing many of the still missing tests, this pull request also changed config.py and classpath.py. In particular, the core functionality of these two modules are wrapped in functions or classes. This was necessary to allow tests (that necessarily have to repeat what was run, already).

src/r5py/r5/transport_network.py
```diff
@@ -99,7 +99,7 @@ def from_directory(cls, path):
                     ),
                     RuntimeWarning,
                 )
-        except KeyError:
+        except IndexError:
```
wklumpen, line 102: What prompted switching from `KeyError` to `IndexError`? Should we include both?

wklumpen (approved):
Left one minor query but overall tests look good.

src/r5py/r5/transport_network.py
christophfink: It was wrong :D and the test showed that. Hooray for test-driven development

christophfink:
For the record: the failing test is a GitHub web server error, it returns http error code 503 (server error) when fetching the r5 jar (which works in all the remaining tests).

PR #233, opened 2022-10-10
[WIP] Wait a bit more for Windows to release file handles

christophfink:
closes #225

PR #236, opened 2022-11-30
bump actions versions

christophfink:


christophfink:
The one test fails for other reasons (re https://github.com/r5py/r5py/pull/233 ), the warnings went away, I’m going to self-approve this one :)

PR #256, opened 2023-02-15
force JVM locale

### LINKED ISSUE #255: ValueError when breakdown=True
Hi,

I'm not managing to get the detailed breakdown of the routing results. I've already tried it out with two different datasets — one being the [Helsinki dataset](https://github.com/r5py/r5py/tree/main/docs/data/Helsinki) from R5py's repo — but every time I get a ValueError. A section of the traceback is attached below.

```
File ~\anaconda3\envs\pyGeo\lib\site-packages\r5py\r5\travel_time_matrix_computer.py:207, in TravelTimeMatrixComputer._parse_results_of_one_origin_details(self, from_id, results)
    196 for ((column_name, column_type), column_data) in zip(
    197     DATA_COLUMNS.items(), zip(*details)  # transpose data
    198 ):

    [...]

    203     # fortunately, we get str, which remains a character string when
    204     # cast to dtype('O')
    206     if isinstance(column_type, pandas.SparseDtype):
    207         column_data = [
    208             [
--> 209                 column_type.subtype.type(value)  # cast
    210                 for value in row.split("|")
    211             ]
    212             if row is not None
    213             else []
    214             for row in column_data
    215         ]
    216     else:
    217         column_data = [
    218             column_type(row) if row is not None else None  # cast
    219             for row in column_data
    220         ]

    [...]

    ValueError: could not convert string to float: '6,0'

```

> **Environment:**
>  - OS: Windows 11 22h2
>  - Package source: conda-forge
>  - Python 3.9.16

All the other examples in the documentation seem to work alright, although there are occasional slight variations in the computed times, but these are probably due to the behaviour described in #240.

It does look like something silly: the decimal separator in the string, I'd say. But I'm not quite sure how to go about it, though: if the problem is on my end, I'm failing to see what's causing it; if it's on your end, I'm not (yet) familiar with the pull request process.

christophfink:
closes #255

### DIFF
--- a/src/r5py/util/jvm.py
+++ b/src/r5py/util/jvm.py
@@ -37,6 +37,9 @@ def start_jvm():
         jpype.startJVM(
             f"-Xmx{MAX_JVM_MEMORY:d}",
             "-Xcheck:jni",
+            "-Duser.language=en",  # setting a default locale, …
+            "-Duser.country=US",  # … as R5 formats numeric return …
+            "-Duser.variant=",  # … values as a localised string
             classpath=[R5_CLASSPATH],
         )
 


christophfink:
Don’t merge yet, this is pending feedback from #255

PR #259, opened 2023-02-15
Add helper method to snap points to network

### LINKED ISSUE #224: Create a helper function similar to r5r’s `find_snap()`
Sometimes, R5 does not manage to properly snap origins or destinations to the network, which results in weird routes or travel times. r5r has a utility function `find_snap()` that ‘simply’ lists all input points, their snapped point and the euclidean distance. This allows the user to then use workarounds, such as assigning the average of neighbouring cells’ travel times to cells with invalid/hugely increased travel times.

cf. https://ipeagit.github.io/r5r/reference/find_snap.html

christophfink:
closes #224


christophfink:
TODO:

This still needs:

- a parameter to `BaseTravelTimeMatrixComputer.__init__()` to automatically snap points, something like `snap_to_network=True` (should the default value be true or false?)
- documentation! (could maybe be done in #238 )

Issue #265, opened 2023-02-25
Combine LegMode, StreetMode, and TransitMode on the user-facing side


wklumpen:
The current version of the documentation is unclear on what a LegMode is vs a TransitMode.

Note that the [documentation on TransitMode
](https://r5py.readthedocs.io/en/stable/reference.html#r5py.TransitMode) contains a series of LegModes. Is `TransitMode` just a collection of these modes? When do you use one or the other?

Suggest changing documentation to spell that out (happy to do it once I fully understand how that works)

christophfink:
Could you maybe check from the R5 source code what the exact differences between LegMode, TransitMode, and StreetMode are? I guess we would find a good place for talking about them more explicitely in the (new, PR #238 , https://r5py--238.org.readthedocs.build/en/238/) documentation

(To be perfectly honest, I remember modelling this after how r5r and/or R5 handle them, and also recall that something about it was counter-intuitive (which one was accepted by which method), but the details evade my memory)

christophfink:
Thinking one step further: what is the use of distinguishing between `TransitMode`, `StreetMode` and `LegMode`? 

If we would combine into one enum (say `TransportMode`), and add code to cast to the required Java types depending on the context, would there be any value lost?

wklumpen:
It appears that r5r just uses strings of the modes (and they are all unique in name e.g. "WALK", "BUS", "TRAM"), which removes the need for the class-based objects. Personally I like that (fits with other packages I'm familiar with). Then we just map it to a dictionary.

christophfink:
This is a great idea, and would go hand-in-hand with the other changes (so: (1) only one user-facing enum, (2) functions accept enum member or string value, (3) enums automatically cast themselves to whatever is needed

wklumpen:
So it appears that "LegMode" refers to the abstract idea of the leg of a journey, while thinks like "Street Mode" and "TransitMode" refer to the actual method of travel - there are conversions between them happening, for example in: https://github.com/conveyal/r5/blob/dev/src/main/java/com/conveyal/r5/api/util/LegMode.java#L29

The question here is should we modify the whole enum assignment structure that's currently set up (which would really just take strings) or just add a layer that accepts a dictionary - the latter would allow users to do either/both.

christophfink:
Let’s also add a check whether destination and origin geometries are within the geometric extent of the GTFS data set. See: https://github.com/r5py/r5py/issues/266#issuecomment-1450134544

[...]

PR #269, opened 2023-03-08
Adding mapping of strings to modes.

wklumpen:
See issue #265.

Also added: tests for string versions of `access_mode`, `egress_mode`, and `transport_mode` specifications

The setters for the various modes now do a quick set comprehension to convert strings into proper enum types using a dictionary added in the `utils` section. Figured that would be the best place for it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ docs/jupyter_execute/
 .tox
 .coverage
 coverage.xml
+.vscode/*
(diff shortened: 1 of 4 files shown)


christophfink:
This definitely works :) Thanks!

(I had imagined to add a constructor to the enums (maybe an intermediate R5pyEnum class or so) that takes either a string or an object of itself, so that we could then do a `StreetMode(mode)` where `isinstance(mode, (StreetMode, str))`. Your solution is definitely more readable, and took less brain-wrecking :+1: )

christophfink (approved):

wklumpen:
I fiddled with some of that and realized perhaps it's easier to not try and be so fancy.

PR #276, opened 2023-03-14
Strict tests of deterministic behaviour [was: Fix random routing results]

### LINKED ISSUE #240: Travel times are non-replicable
**Describe the bug**
Given the same input data (origins/destinations, OSM, and GTFS), and the same parameters, r5py will return travel-time matrices with very significant differences. This seems to happen whether or not I rebuild the network or regenerate the computer in between runs. Most O/D pairs are the same, but a small fraction of pairs are substantially different - off by 20% or more of the travel time, often tens of minutes, enough to raise serious doubts concerning the quality of the results. 

I'm not an expert, it's very likely that I've made some mistake in configuring my inputs or my parameters, but I've been digging through the r5py documentation and trying everything I can think of, and I can't get r5py to give replicable results. Help me find what I'm doing wrong?

**To Reproduce**
0. I've tested both the development version of r5py and the stable release, they seem to have the same issue.
1. Use this [Google Drive](https://drive.google.com/drive/folders/1RsCRm7ejjTZ9kshMO7sg3_G5qbhwBFsh?usp=sharing) to download the data (scenarios/existing_conditions/) and the script (routing.py) that I'm using to generate travel_time matrices.
2. Run routing.route_scenario('scenarios/existing_conditions/'), and move the resulting files out of the results/ directory.
3. Run routing.route_scenario('scenarios/existing_conditions/') again, and compare the results to the first run. 

**Expected behavior**
For my intended use in transportation planning, I need a routing engine that can be configured to return identical, deterministic results every time an analysis is run on the same input data with the same parameters.

**Environment:**

    OS: Ubuntu 20.04.4 LTS
    Python package source: conda
    Versions of Python, Java Development Kit, Python modules
    Python 3.9
    javac 11.0.16 (I think, from javac -version, I'm not certain how to check within conda/python)

christophfink:
closes #240 (for good, this time)


src/r5py/r5/regional_task.py
wklumpen, line 1: Is this `monteCarloDraws` value not needed for headway sampling?

christophfink: They way I read R5’s code, the monte carlo draws are only used with frequency-based schedules; otherwise, every minute of departure_time_window is evaluated exactly once

See https://github.com/conveyal/r5/blob/dev/src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java#L153

christophfink: But yes: should we use 1 monte carlo draw as a default? (https://github.com/conveyal/r5/blob/dev/src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java#L158)

christophfink: >             // In Monte Carlo mode, each iteration is a fresh randomization of frequency route offsets.
            // In half-headway mode, only one iteration will happen and schedules will not be randomized.

christophfink: Sorry for spamming: this needs careful reading of https://github.com/conveyal/r5/blob/dev/src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java - there’s quite a bit of conditional logic in there that’s not immediately accessible on first read

wklumpen: My main inquiry was to make sure that setting the value to 0 wouldn't cause issues with headway-based sampling (where I think an MC approach makes sense).  Does setting this value to > 0 cause more randomness in non-headway schedules? If so it would indeed be good to understand what's going on under the hood.

christophfink: Yeah, I’m also lacking the time, I’ll try to come back to it later this week

christophfink:
```
    /**
     * The different ways of determining the wait time before boarding a vehicle for frequency-based routes.
     * Would it be more efficient to pass in a function reference, after defining a new departure search interface?
     * e.g. public interface frequencyDepartCalculator { ... }
     */
    public enum FrequencyBoardingMode {
        /** The primary frequency search, using various randomized departure offsets for each route. */
        MONTE_CARLO,
        /** Compute a deterministic upper bound, which helps speed up subsequent frequency searches. */
        UPPER_BOUND,
        /** Assume a vehicle always comes after exactly half the headway, for comparison with Monte Carlo. */
        HALF_HEADWAY
    }
```

christophfink: https://github.com/conveyal/r5/blob/dev/src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java#LL583C1-L595C6

christophfink: The way I read it, the Monte Carlo draws are only used with frequency-based schedules

christophfink: Which means this merge request should be changed so the value adjusts depending on whether or not we are using frequency-based GTFS data. In a first, easiest, step, I will disable monte carlo draws for all other transport modes than public transport (where it weirdly seems to have an influence)

christophfink: @wklumpen : you were right, the `monteCarloDraws` are indeed also used with time-based schedules, not only with frequency-based ones. The value defines how many departure times are chosen ‘randomly’ in the departure time window. ‘Randomly’ left in quotation marks, because the randomisation is seeded with the origin coordinate values to ensure repeatability. I reverted the `request.monteCarloDraws` value to `60` (what r5r uses). With other changes implemented here and in other pull requests, we got rid of the randomness, and the now strict tests in `test_deterministic_behaviour.py` seem to pass (offline, let’s see what the GitHub action run says 🤞)

src/r5py/r5/transport_network.py
```diff
@@ -125,7 +125,7 @@ def __del__(self):
             if not temporary_files:  # empty
                 break
 
-            print(f"waiting for still open files, round {_}")
+            # there are still files open, let’s wait a moment and try again
```
christophfink, line 128: This is only cleaning a left-over debug statement

tests/test_deterministic_behaviour.py
```diff
@@ -65,22 +65,8 @@ def test_public_transport_travel_time_matrices(
...
-                m_a.diff_rel.describe(),
-                # m_a[(m_a.difference >= 3) | (m_a.difference <= -3)],
-            )
-
             pandas.testing.assert_frame_equal(
                 matrix_a,
                 matrix_b,
                 check_like=True,
-                atol=4,  # tolerance between runs, in minutes
             )
```
christophfink, line 72: And this removes the tolerance when comparing two consecutive runs of the same analysis: the resulting values now have to be 100% identical

PR #284, opened 2023-04-11
Freeze Binder environment at Python 3.10

christophfink:


christophfink:
I merged the dev env.yaml from the upcoming documentation update, that’s where the additional dependencies come from

wklumpen (approved):

PR #286, opened 2023-04-14
Simplify transport modes

christophfink:
This is a first step for implementing a different DetailedItinerariesComputer. It’s a separate PR to ease running the tests.


christophfink:
I created a separate issue for updating the respective documentation, see #290

HTenkanen:
@christophfink Is this ready for review? I see lots of things happening so just wanted to make sure before diving in? 🙂

tests/test_transport_mode.py
HTenkanen, line 1: Should this one also test the case if user passes incorrect TransportMode? Like typos (CRA) or modes that are not supported (helicopter).

HTenkanen (approved):
One small thing to consider in tests, but otherwise looks good to me! I like this logic of reducing a bit the complexity of r5 from the user 👍🏻

tests/test_transport_mode.py
christophfink: Good point! I added such tests. Beyond that, I also made sure the case of a transport mode passed in as `str` does not make a difference, and added tests to check that that works.

christophfink:
@HTenkanen : just pushed another small commit to address the concern you brought up, could you take another look?

PR #299, opened 2023-07-11
Update Binder environment to Python 3.11

christophfink:


christophfink:
Test here: https://mybinder.org/v2/gh/christophfink/r5py/gesis-py311

PR #318, opened 2023-08-11
Handle `origins == destinations` cases in `DetailedItinerariesComputer`

### LINKED ISSUE #312: Warn if `origins.equals(destinations)` in `DetailedItinerariesComputer.__init__()` (same-to-same routes)
**Describe the bug**
As titled, the `DetailedItinerariesComputer` appears to not compute segments/travel times for a Calgary example, when `TravelTimeMatrixComputer` produces realistic results.

**To Reproduce**
Steps to reproduce the behaviour:
1. Download the attached zip data
2. Run the included Python workbook using an environment

**Expected behaviour**
A detailed itineraries list that mimics the travel time matrix produced via the `TravelTimeMatrixComputer`

**Environment:**
 - OS: Ubuntu 22.04.2 LTS
[Expanded Matrix Example.pdf](https://github.com/r5py/r5py/files/12314835/Expanded.Matrix.Example.pdf)
[calgary_data.zip](https://github.com/r5py/r5py/files/12314894/calgary_data.zip)

christophfink:
closes #312

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,6 +20,7 @@ dependencies = [
     "filelock",
     "fiona",
     "geopandas",
+    "importlib_resources",
     "joblib",
     "jpype1",
     "numpy",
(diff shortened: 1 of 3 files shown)


wklumpen (approved):
Looks good; I'm starting to like this "do something intuitive for the user but print a warning" approach.

PR #362, opened 2023-09-25
[WIP] Added agency and stop info to detailed itineraries

wklumpen:
This PR adds more detail to the transit portions of the detailed itineraries.

The main thing left to do here is generate the updated data column. @christophfink how did you go about that? It looks like a lot of the data is trimmed in a certain way; do you have a pattern for regenerating the test data?


wklumpen:
Need to add to the Detailed Itineraries docs and figure out why tests are failing here but not on my local machine

PR #368, opened 2023-10-16
Catch Java exception in destructor

christophfink:


christophfink:
This is now ready for review

Note that the changes got a bit bigger than anticipated: the actual fix for the issue is the change in https://github.com/r5py/r5py/pull/368/files#diff-443d8d96b5a3ee79f8fe4516b383ae7826bbc38932f3c37de72d1260e717ce4fR98 , but with pandas, geopandas, and numpy having breaking updates upstream, a few extra changes were necessary, including retiring our support for Python 3.8

Issue #366, opened 2023-10-12
java.lang.RuntimeException after completing the travel time calculation 


shiqin-liu:
**Describe the bug**
I imported R5py as a Python module in a script to run the travel time calculation. I get this Java error at the end every time I complete the calculation. Look like Java stops running when R5py attempts to delete some files. It does not seem to affect the overall calculation process and the result outputs though.

```
Exception in thread "MapDB shutdown hook" java.lang.RuntimeException: java.lang.InterruptedException
	at org.mapdb.AsyncWriteEngine.close(AsyncWriteEngine.java:443)
	at org.mapdb.EngineWrapper.close(EngineWrapper.java:81)
	at org.mapdb.Caches$HashTable.close(Caches.java:328)
	at org.mapdb.EngineWrapper.close(EngineWrapper.java:81)
	at org.mapdb.EngineWrapper$CloseOnJVMShutdown.close(EngineWrapper.java:586)
	at org.mapdb.EngineWrapper$CloseOnJVMShutdown$1.run(EngineWrapper.java:571)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.lang.InterruptedException
	at java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedNanos(AbstractQueuedSynchronizer.java:1081)
	at java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer.tryAcquireSharedNanos(AbstractQueuedSynchronizer.java:1369)
	at java.base/java.util.concurrent.CountDownLatch.await(CountDownLatch.java:278)
	at org.mapdb.AsyncWriteEngine.close(AsyncWriteEngine.java:435)
	... 6 more
Exception ignored in: <function TransportNetwork.__del__ at 0x14b3dbd90>
Traceback (most recent call last):
  File "/Users/ssliu/.pyenv/versions/nae/lib/python3.10/site-packages/r5py/r5/transport_network.py", line 98, in __del__
jpype._core.JVMNotRunning: Java Virtual Machine is not running
```

**Environment:**
 - MacOS
 - Python package source (PyPi)
 - Python 3.10, Java Development Kit 11

Any insight on this would be appreciated!

christophfink:
[...]

PR #375, opened 2023-10-23
Fix typehints for `TransportNetwork.__init__()`

christophfink:
I’ll still merge #366 into here to pass the tests

Be sure to accept #366 first


christophfink:
Still needs tests whether arguments in all possible combinations of types work

Issue #382, opened 2023-11-17
Improve error handling on loading of GTFS


wklumpen:
**Is your feature request related to a problem? Please describe.**
I've come across a few times where I've had an issue loading a GTFS dataset during the transport network building phase.

**Describe the solution you'd like**
Let's place a try/except block around the following block:

```python
gtfs_feed = com.conveyal.gtfs.GTFSFeed.readOnlyTempFileFromGtfs(gtfs_file)(gtfs_file)
transport_network.transitLayer.loadFromGtfs(gtfs_feed)
gtfs_feed.close()
```
with an error message that prints the feed that's failed to load.

**Describe alternatives you've considered**
None - this is a clarity and additional QOL issue.

**Additional context**
It would be great to know what part of the feed is throwing the issue, but that might be more than we're advertising for this.

jcorral00:
Hi! 

I'm encountering an issue when trying to load GTFS datasets during the transport network building phase using r5py. Despite ensuring that file paths are correct and that I have the necessary file permissions, I consistently receive the error message Error loading GTFS.

Here are the specifics of my environment and the error:

- Operating System: Windows 11 Home x64
- Python Version: 3.10 
- Java Version: 11.0.22
- r5py Version: 0.1.1

`Error loading GTFS feed: [Path to the GTFS zip]
Detail of the error: com.conveyal.gtfs.GtfsLibException: Error loading GTFS.
`
This issue persists even after addressing all warnings from the GTFS feed validation (such as expired calendars, missing recommended columns, etc.). It's unclear which part of the GTFS feed is causing the issue, as no further details are provided in the error message.

Could there be any compatibility issues with the versions of the tools I'm using, or might there be some other underlying cause that I'm overlooking? Any assistance or insight into resolving this error would be greatly appreciated.

Thank you for your time and help.

wklumpen:
Hi there!

Have you tried running your GTFS feed through the [MobilityData Canonical Validator](https://gtfs-validator.mobilitydata.org/) to see if there are any systemic issues?

My advice is to clean your GTFS feed as much as possible down to the barebones data. GTFS-Lite can help with that.

PR #383, opened 2023-11-17
[WIP] Added try/except block to GTFS loading.

wklumpen:
See #382


christophfink:
This PR has gone stale, see #471 instead

PR #385, opened 2023-11-22
Use our own R5 jar, upgrade R5 to v7.0

### LINKED ISSUE #308: Provide our own R5 jar that is compiled with `TransitLayer.SAVE_SHAPES = true`
`DetailedItinerariesComputer`, with the default upstream R5 jar, cannot report the geometries of public transport routes. A simple one-line patch to R5 would suffice.

I suggest to fork R5 within the r5py GitHub organisation, set up a GitHub actions toolchain to compile patched versions of the upstream releases, and use those as the default jars downloaded by r5py

### LINKED ISSUE #384: support R5 7.0
Upstream just released a new major version

https://github.com/conveyal/r5/compare/v6.9..v7.0

christophfink:
closes #308 
closes #384


wklumpen (approved):
I like your approach better than mine on the future warning

christophfink:
> I like your approach better than mine on the future warning

It looks more elegant, but maybe not as robust: I’m still wondering what happens if the first data frame returned is empty - not sure that will ever happen, though

christophfink:
RTD is failing as Ubuntu does not (yet) have Java 21 🤯

cf. https://askubuntu.com/a/1492589

christophfink:
Seems this will take (at least) about a week to be in the normal package sources from which readthedocs can install: https://bugs.launchpad.net/ubuntu/+source/openjdk-21/+bug/2036873

christophfink:
For the record: as it seems to take a while until Java-21 drops to Ubuntu-22.04, I instead changed readthedocs to use a conda environment. On the way I discovered that conda does not always export all environment variables properly (`JAVA_HOME`, `PROJ_LIB`), and wrote a workaround for those cases

HTenkanen:
> For the record: as it seems to take a while until Java-21 drops to Ubuntu-22.04, I instead changed readthedocs to use a conda environment. On the way I discovered that conda does not always export all environment variables properly (`JAVA_HOME`, `PROJ_LIB`), and wrote a workaround for those cases

@christophfink: Sorry, I most likely have missed or forgotten discussion about this, but: Wouldn't it be much simpler to NOT run the Notebooks  on ReadTheDocs by specifying `jupyter_execute_notebooks = "off"` for Sphinx in `conf.py`? Or was there a specific reason why we want to always run them there? If we wouldn't run the Notebooks, things would be very simple with RTD.

HTenkanen (approved):

HTenkanen:
> Then we have to push the executed notebooks that change all the time. Having the input notebooks in git is so much cleaner

Ah right, now I remember. Fair enough and happy to continue like this 👍🏻 All looking good!

Issue #387, opened 2023-12-05
Are origin threads expected to be run in parallel?


owenam:
**Describe the bug**
Based on the threading setup in [TravelTimeMatrixComputer.compute_travel_times()](https://github.com/r5py/r5py/blob/18d8bd4e04038af25cd88449b680c9d705c3b2d5/src/r5py/r5/travel_time_matrix_computer.py#L25), I expected matrices for multiple origins to be calculated in parallel. Maybe this was a bad assumption? Matrix calculation times do not seem affected by the number of threads allocated. 

**To Reproduce**
I'm calling setting up a `TravelTimeMatrixComputer` with constant parameters, explicitly setting `NUM_THREADS`. This is on an AWS EC2 instance with 32 logical cores and 64 GiB RAM (c7g.8xlarge). The input network for this test is a slice of rural MN, using only OSM data (no GTFS), though I am seeing the same behavior for any OSM extract and both with and without GTFS.
```python
computer = r5py.TravelTimeMatrixComputer(**ttm_parameters)
computer.NUM_THREADS = 1
print(f"Running on {computer.NUM_THREADS} threads")
df = computer.compute_travel_times().dropna()
print(f"Completed {mode}/{lts} calculations in {t.elapsed()}")
```

Regardless of the value of `NUM_THREADS`, compute_travel_times() returns in approximately the same time:
##### MAX_THREADS = 1
```
Calculating travel time matrix for walk/None...
Running on 1 threads
Completed walk/None calculations in 0:01:12.273547
```

##### MAX_THREADS = 2
```
Calculating travel time matrix for walk/None...
Running on 2 threads
Completed walk/None calculations in 0:01:07.383546
```

##### MAX_THREADS = 4
```
Calculating travel time matrix for walk/None...
Running on 4 threads
Completed walk/None calculations in 0:01:09.598225
```

##### MAX_THREADS = 8
```
Calculating travel time matrix for walk/None...
Running on 8 threads
Completed walk/None calculations in 0:01:15.273143
```

##### MAX_THREADS = 16
```
Calculating travel time matrix for walk/None...
Running on 16 threads
Completed walk/None calculations in 0:01:11.566810
```

**Expected behavior**
I expected that multithreading would allow multiple origins to be run in parallel, providing some degree of speedup vs single-threaded operation. Maybe I'm doing something wrong, or maybe that's not the intended behavior? If it's not, then what does the multithreading in `compute_travel_times()` provide?

**Environment:**
[...]

PR #392, opened 2024-01-15
Leave parallelisation to Java

christophfink:
As described in detail in #387 , running the per-origin travel time computations in parallel does not, in fact, speed up computation (it might for *very* small data sets)


ctriley:
@christophfink quick question about this -- how do we get java to parallelize? Right now (at least the way I'm using r5py) it is only using 1 core.

PR #401, opened 2024-02-12
Multi-objective routing / custom cost edge weight factors

roopehub:
Here's a overview of changes made to files

advanced-use.md
- added  documentation to all Multi-objective routing changes

custom_cost_transport_network
- added a class which extends TransportNetwork with "custom cost" functionality.

direct_leg
- check if street_segment.StreetEdges have "edgeOsmId" and if so, add those to super initialization

transport_network
- add check if using CustomCostTransportNetwork, if so, inject the costFields to the network.
- add conditionals to only execute the gtfs/public transport related codes if gtfs is provided. 

travel_time_matrix_computer
- check if results have osmIdResults, if so, add them to the result GDF as new column

trip_leg
- add support for osmids

trip_planner
- add support for osmids 

custom_cost_conversions
- handle converting from python to java and vice versa
- build CustomCostFields in here

exceptions
- add new custom exceptions for multi-objective routing

tests and conftest
- add datas for testing (quite mig csv?)
- add tests for custom cost routing / multi-objective routing


christophfink:
Hi Roope! Thanks for this PR, looks exciting, but will take a while to review I guess :)

Could you run the  `flake8` linter or otherwise adapt your code to pass (see the output in the failed workflow run: https://github.com/r5py/r5py/actions/runs/7872801098/job/21478795874?pr=401)?

Once the tests run through, I can then adapt your RTD notebook to use your custom, patched R5, as this is another workflow that currently fails: https://readthedocs.org/projects/r5py/builds/23416925/

Overall, this looks very promising, thanks for your hard work!

Chris

roopehub:
I ran the flake8, linting should work now :) @christophfink

.gitignore
```diff
@@ -39,3 +39,7 @@ coverage.xml
 .vscode/*
 
 .config/r5py.yml
+*.ipynb
```
christophfink, line 42: please don’t exclude all ipynb and all csv files

docs/user-guide/user-manual/advanced-use.md
christophfink, line 1: Could you maybe move this to a separate page, not into Advanced Use, as it’s four times longer than the original page. So maybe you could have another file `multi-objective-routing.md` that then is linked before `advanced-use.md`?

src/r5py/r5/custom_cost_transport_network.py
```diff
@@ -0,0 +1,399 @@
...
+)
+from r5py.util.exceptions import CustomCostDataError
+from r5py.util.jvm import start_jvm
+
+__all__ = ["CustomCostTransportNetwork"]
+
+start_jvm()
+
+
+def r5_supports_custom_costs():
```
christophfink, line 18: Could this function go into the `r5.utils` somewhere? (I saw you use it outside of this file, as well, and it would be nice to not have a global function in `__all__`, here)

christophfink: Or, alternatively, add this as a class method to `CustomCostTransportNetwork`

src/r5py/r5/custom_cost_transport_network.py
```diff
@@ -0,0 +1,399 @@
...
+            Define whether to allow missing osmids.
+
+        Returns:
+        --------
+        names : List[str]
+        sensitivities : List[float | int]
+        custom_cost_segment_weight_factors : List[Dict[str, float]]
+        allow_missing_osmids : List[bool]
+        """
+        if not isinstance(names, list):
```
christophfink, line 143: This does not work if, e.g., a tuple or another Iterable other than list is passed. Use `isinstance(maybe_a_list, collections.abc.Iterable)`

christophfink: I assume that it’s not necessary that the parameters are lists, specifically, but just behave like lists?=

.gitignore
roopehub: true good point, will remote this

docs/user-guide/user-manual/advanced-use.md
roopehub: good point, moved to the multi-objective-routing.md and added link to toctree in quickstart.md

src/r5py/r5/custom_cost_transport_network.py
roopehub: good idea, moved to r5py.utils.classpath, hopefully that is suitable place for it, the file has "find_r5_classpath" so sounds that it handles similar logic

src/r5py/r5/custom_cost_transport_network.py
```diff
@@ -0,0 +1,399 @@
...
+    def validate_custom_cost_params(
+        self,
+        names,
+        sensitivities,
+        custom_cost_segment_weight_factors,
+        allow_missing_osmids,
+    ):
+        """
+        Validate CustomCostTransportNetwork parameters.
+        All parameters are transformed into lists already in convert_params_to_lists method.
```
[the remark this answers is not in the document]
roopehub: good idea, moved list conversion from __init__ to here

src/r5py/r5/custom_cost_transport_network.py
roopehub: that's true and smart, i added Iterable support for params and some additional conditionals for e.g. string and dict (?) are iterable so they needed still checks not to pass faulty

src/r5py/r5/custom_cost_transport_network.py
```diff
@@ -0,0 +1,399 @@
...
+                    custom_cost_segment_weight_factors
+                )
+        return (
+            names,
+            sensitivities,
+            custom_cost_segment_weight_factors,
+            allow_missing_osmids,
+        )
+
+    def validate_custom_cost_params(
```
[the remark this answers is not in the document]
roopehub: made it "private" with underscore

PR #416, opened 2024-06-03
Test with  numpy>=2.0

christophfink:


christophfink:
Seems this works! We’re prepared for numpy 2.0

PR #427, opened 2024-07-26
split conftest into multiple files

christophfink:


christophfink:
I inadvertently reverted the changes from [PR #424](https://github.com/r5py/r5py/pull/424) 
They still need to copied over

PR #444, opened 2024-10-05
Test with Python 3.13, upgrade PyPi upload action, migrate to TrustedPublisher 

christophfink:


christophfink:
This is on hold for the time being, as `pyarrow` does not provide a wheel for 3.13 yet and also does not successfully build in a pip/pep531 installation procedure.

PR #472, opened 2025-04-02
Elevation loading

christophfink:


christophfink:
R5‘s raster loading code does not understand all valid formats of tiffs, we will have to convert to some easy-peasy vanilla file (uncompressed works, but that‘s not feasible - how much cache should we have!?)

christophfink:
> The failed tests are `shapely` having been updated to `2.1.0` in the middle of the test runs. The new version raises `DeprecationWarning`s for using `shapely.geos` - current `geopandas` imports `shapely.geos`

Note that this has been fixed in trunk geopandas, let’s hope there’s a release, soon

PR #512, opened 2026-01-15
Fix DataFrame slicing by overriding _constructor in BaseTravelTimeMatrix

ibarapascal:
## Description

DataFrame operations (slicing, filtering, `dropna()`) on `TravelTimeMatrix` and other `BaseTravelTimeMatrix` subclasses fail with `FileNotFoundError`.

- **Goal**: Fix DataFrame slicing operations on `BaseTravelTimeMatrix` subclasses
- **Affected code**: `src/r5py/r5/base_travel_time_matrix.py`

**Root cause**: `BaseTravelTimeMatrix` extends `GeoDataFrame` but does not override `_constructor`. When pandas performs slicing, it calls `_constructor` to create a new instance, which invokes `__init__()` with pandas internal data manager objects instead of a `TransportNetwork`, causing unexpected file access errors.

**Solution**: Override `_constructor` to return `GeoDataFrame` and `_constructor_sliced` to return `Series`. This is the standard pattern for GeoDataFrame subclasses that shouldn't propagate their special behavior through pandas operations.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] I have read the [CONTRIBUTING guide](https://r5py.readthedocs.io/en/stable/contributing/CONTRIBUTING.html).
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have added or updated documentation where necessary, and ensured it fulfills the `pydocstyle` standards.
- [x] I have formatted my code with `black` and ensured linting passes (`flake8`).
- [x] I have made sure my code follows the style guidelines of this project.
- [x] I have rebased/merged the latest changes from `main`.
- [ ] I have verified that CI passes after my changes.

## Screenshots / output (if applicable)

**Before fix:**
```
FileNotFoundError: [Errno 2] No such file or directory: '~/.cache/r5py/from_id'
```

**After fix:**
```python
filtered = ttm[ttm['travel_time'] <= 30]  # works correctly
```

## Additional notes

Affected classes (all inherit from `BaseTravelTimeMatrix`):
- `TravelTimeMatrix`
- `TravelTimeMatrixComputer` (deprecated)
- `Isochrones`
- `DetailedItineraries`

### DIFF
--- a/src/r5py/r5/base_travel_time_matrix.py
+++ b/src/r5py/r5/base_travel_time_matrix.py
@@ -8,7 +8,6 @@ import warnings
 
 import geopandas
 import numpy
-import pandas
 import shapely
 
 from ..util import check_od_data_set, Config
@@ -52,11 +51,6 @@ class BaseTravelTimeMatrix(geopandas.GeoDataFrame):
     def _constructor(self):
         """Return GeoDataFrame for slicing operations."""
         return geopandas.GeoDataFrame
-
-    @property
-    def _constructor_sliced(self):
-        """Return Series for single-column selection."""
-        return pandas.Series
 
     def __init__(
         self,


src/r5py/r5/base_travel_time_matrix.py
```diff
@@ -47,6 +48,16 @@ class BaseTravelTimeMatrix(geopandas.GeoDataFrame):
         "verbose",
     ]
 
+    @property
+    def _constructor(self):
+        """Return GeoDataFrame for slicing operations."""
+        return geopandas.GeoDataFrame
```
christophfink: why not `pandas.DataFrame`?

Issue #466, opened 2025-03-13
`RegionalTask.breakdown` does not work as intended, anymore


christophfink:
Upstream changed PathResult.MAX_PATH_DESTINATIONS to final, so this does not work for destinations with len()>5000

see https://github.com/conveyal/r5/commit/2481a931e25279cfb5ed606c038a7dd4b5992dd8

Not quite sure whether breakdown has worked recently, anyway (it was a hidden feature that we did not test really, either)

christophfink:
To clarify, `breakdown=True` has not had an effect for a while, let's clean up the code and remove it

Issue #305, opened 2023-08-03
Report wait time for public transport results from `TravelTimeMatrixComputer`


christophfink:
We’ve run into odd artefacts in our travel time matrices for Helsinki with public transport routes during nighttime. Specifically, when plotting histograms of `travel_time`, we notice odd steps that seem to relate to the fact that `wait_time` is not included in what is currently reported by `TravelTimeMatrixComputer.compute_travel_times()`. Departure time is not either known.

This is of course far away from the experience of a traveller: they will, indeed, care about having to wait for an hour before being able to catch a ride, even if that ride is very quick. On top of that, it is also slightly inconsistent when using longer departure time windows: from the last minute of such a window, the router might decide to still wait up to `max_wait_time`, resulting in a departure time much later than expected.

To resolve this issue, I suggest to include a boolean argument `include_wait_time` (default: `True`) to `RegionalTask` that causes `TravelTimeMatrixComputer` to report as `travel_time` the sum of waiting and travel time. For more detailed breakdowns, users can use `DetailedItinerariesComputer`s.

christophfink:
Alright. After a bit of experimenting (see https://github.com/christophfink/r5py/tree/305-wait-time ), I’ve learnt that the normal `travel_time` already includes the wait time (a reasonable default, IMHO), but that we, so far, lack means to restrict the maximum waiting time. 

What happens in our example is the following:

- we search for PT connections in a departure window between 2 and 3 am
- the first bus connection to a certain place leaves at 4:21am from an intermediate stop that can be reached by departing within the departure window (on a different ride)
- R5 finds that connection, reports a total travel time (aggregated from the values of each of the iterations, i.e., including waiting time from many different points in time within the departure window)
- Much of the accrued waiting time is in-between rides: departure is within the time window, but the first bus onwards from an intermediate stop is much later, leading to characteristic step in a histogram plot of travel times (by origin, to a common destination)

In a way, this is okay from the point of view of the imaginary traveller: if they want to travel from A to B, leaving in a certain time window, their (median, or at another percentile) travel time includes the waiting time. 

I think I’ll close this issue for now. There is one aspect of it that would be interesting, though, and that the code I wrote so far would already mostly cover: for some use cases it would interesting to know travel times without the wait time before the first ride, i.e., without the ‘waiting at home’ part. (However, this can be achieved by requesting the 0-th percentile, at least for departures within the departure time window)

PR #547, opened 2026-08-26
Reintroduce travel time breakdown as TravelTimeMatrixDecomposed

### LINKED ISSUE #543: Travel time breakdown (in-vehicle, wait, access, egress)
For my work in travel-demand modeling, I need a transit routing option that's capable of returning results on the composition of a given trip: most importantly, the breakdown of time spent in- vs out-of-vehicle, ideally including the breakdown of time spent walking vs waiting and the number of transfers. 

R5 computes exactly this breakdown already and r5py used to expose it, so I would like to propose bringing it back. I am happy to write the code and submit a PR.

## What happened to `breakdown`

Tracing the history, the feature may have been lost by accident:

These stats were once collected by `TravelTimeMatrixComputer._parse_results_of_one_origin_details()`, which parsed the strings from `PathResult.summarizeIterations()`.
[#433](https://github.com/r5py/r5py/pull/433), the refactor to DataFrame-inheriting classes, deleted `travel_time_matrix_computer.py`. The new `TravelTimeMatrix._parse_results()` reads only `results.travelTimes`, so `OneOriginResult.paths` stopped being read. The breakdown property in RegionalTask survived and kept setting `includePathResults`, but nothing consumed the output any more. I believe this is what [#466](https://github.com/r5py/r5py/issues/466) observed as "has not had an effect for a while." [#467](https://github.com/r5py/r5py/pull/467) then removed the orphaned flag, along with `BreakdownStat`.

So the R5 side was never the problem. With `includePathResults` set, R5 still computes and returns the full path breakdown on every call.

## Proposal
I propose to collect the breakdown statistics by reading R5's public `PathResult.iterationsForPathTemplates` array directly, indexed by destination, and pulling the numeric components off the `RouteSequence`/`StopSequence`/`Iteration `objects (in-vehicle from `stopSequence.rideTimesSeconds`, access/egress from `stopSequence.access/egress`, wait from `iteration.waitTimes`, transfer from `stopSequence.transferTime(iteration)`, total from `iteration.totalTime`). 

This avoids R5's string-formatting path entirely, which sidesteps two traps: the locale-dependent decimal separator behind [#255](https://github.com/r5py/r5py/issues/255) ('6,0'), and a NullPointerException that R5's `summarizeIterations()` throws on transit-leg paths when `csvResultOptions` is null. I verified this object-graph read against the many-to-many matrix case: components reconcile to `totalTime` exactly (max abs residual 0.0000 min over 6,281 iterations), and R5's own single-destination accessor `getPathIterationsForDestination()` is not usable here because it asserts a single destination.

On the destination cap from [#466](https://github.com/r5py/r5py/issues/466), I would not try to restore the patching of PathResult.MAX_PATH_DESTINATIONS now that it is final. Instead: validate up front and raise a clear r5py exception above 5000. We may want to raise a follow-up issue to additionally batch destinations in chunks of 5000. 

Tests and user guide documentation will be part of the PR.

### 1: Statistical inconsistency

`travel_time` is the median (or arbitrary percentile) over the departure time window, while the path components summarize as `Stat.MEAN` or `Stat.MINIMUM` over iterations. These are different statistics, so the components will not sum to the reported travel_time for a given OD pair, and it gets worse if non-default percentiles are requested. 

There is also a selection effect — if some OD pairs are not reached in all 60 draws, their component means are conditional on the destination being reachable at all, which understates wait time specifically. Related: path results may cover more OD pairs than the matrix reports as reachable, and some percentage of returned path rows are walk-only with no transit leg. None of this is a blocker, but the API has to either expose or clearly document it, and I would rather agree on which up front.

#### My proposal for statistical inconsistency

Forward R5's values as reported, at R5's native granularity, with explicit labels and enough context columns that the user can aggregate, weight, or filter however they need.

This will make the user responsible for interpretation. There are some risks with this -- new users might mistakenly misinterpret results, and the misalignment with p50 travel-time results might cause confusion. But I think this approach will provide maximum freedom for users who want to include these statistics in a variety of methodologies. And since this will be an optional function anyway, I think the risk is worth the payoff.

Specifically, I'd return the breakdown as its own long-format table, separate from the wide travel_time matrix — not as extra columns bolted onto the matrix row. Granularity: one row per (origin, destination, path template), which is exactly what R5's PathResult gives — a distinct row for each RouteSequence R5 found between the pair. Columns:

* from_id, to_id
* routes, board_stops, alight_stops, feed_ids — the path identity, as R5 reports it
* in_vehicle_time, wait_time — per-leg (R5 reports these as per-leg lists; keep them as lists rather than pre-summing, so a user who wants per-leg wait, e.g. to weight transfer wait differently from initial wait, still can)
* access_time, egress_time, transfer_time, total_time — scalar, as R5 reports
* n_rides -- derived from above, presented for convenience
* n_iterations — how many departure-minute draws used this template

The per-template summary statistic (mean/minimum) will be exposed as a parameter, reintroducing BreakdownStat, and recorded on the output.

### 2: API shape. 

Since [#337](https://github.com/r5py/r5py/issues/337) moved to noun classes, a sibling class alongside TravelTimeMatrix and DetailedItineraries may be more idiomatic than reviving a boolean argument, but r5r's precedent is expanded_travel_time_matrix() with breakdown=TRUE and there is an argument for matching it. I have no strong preference. Without guidance, I'd probably implement a new sibling to `TravelTimeMatrix`, something like `TravelTimeMatrixDecomposed` or `ExpandedTravelTimeMatrix` (the latter matches r5r's nomenclature but has slightly different behavior). 

This would also substantially cover [#305](https://github.com/r5py/r5py/issues/305), which proposed reporting wait time and was closed without it.

(Issue prepared with use of Claude Code / Opus 4.8. I plan to use the same LLM in writing the code.)

dtreich:
## Description

This implements the proposal in #543: it brings back the travel time breakdown that was lost in the #433 refactor, as a new class alongside `TravelTimeMatrix`.

- `TravelTimeMatrixDecomposed` returns a long-format table with one row per (origin, destination, path template): the total travel time split into in-vehicle, waiting, access, egress, and transfer components, together with the path identity (routes, their GTFS `route_type`s, board/alight stops, feed ids) and the number of departure-minute iterations the template served. Per-leg `in_vehicle_time` and `wait_time` are kept as lists rather than pre-summed, so users can weight (e.g.) initial vs. transfer waits differently.
- Components are read from R5's public object graph (`PathResult.iterationsForPathTemplates`) rather than its CSV string formatting, which sidesteps the locale-dependent decimal separator behind #255 and a `NullPointerException` that `summarizeIterations()` throws when `csvResultOptions` is null.
- `BreakdownStat` (`MEAN` / `MINIMUM`) is reintroduced to choose the representative iteration per template, mirroring R5's own summarization logic.
- Destination counts above R5's `PathResult.MAX_PATH_DESTINATIONS` (5000, now final) are rejected up front with a clear error; batching larger requests could be a follow-up issue (context in #466).
- The statistical caveats discussed in #543 — component statistics are not the same statistic as the percentile travel time, path templates are conditional on reachability, walk-only rows appear with `n_rides == 0` — are documented in the new user-guide chapter rather than hidden.

New files only; no existing classes change: `travel_time_matrix_decomposed.py`, `breakdown_stat.py`, a test module, a user-guide chapter, reference-docs and changelog entries, and the package exports.

Fixes #543. Substantially covers #305.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have read the [CONTRIBUTING guide](https://r5py.readthedocs.io/en/stable/contributing/CONTRIBUTING.html).
- [x] I have added tests that prove my feature works.
- [x] I have added documentation and ensured it fulfills the `pydocstyle` standards.
- [x] I have formatted my code with `black` and ensured `flake8` passes.
- [x] I have made sure my code follows the style guidelines of this project.
- [x] I have rebased the latest changes from `main` (including adapting to R5 v7.6's `StopSequence` API).
- [ ] I have verified that CI passes after my changes. *(will confirm once CI runs on this PR)*

## Screenshots / output

Helsinki sample data, one origin, 60-minute departure window (note the two-leg row: per-leg lists, extended GTFS route types, non-zero transfer time):

```
from_id to_id     routes route_types     board_stops    alight_stops  in_vehicle_time              wait_time  access_time  egress_time  transfer_time  total_time  n_rides  n_iterations
      0     2       1008           0         1201402         1203401            [1.0]                 [1.37]         3.63         4.40           0.00       10.40        1             6
      0    10       1008           0         1201403         1130442            [9.0]                 [4.13]         2.87         6.60           0.00       22.60        1            13
      0    10 7192T|1020     700|702 1201132|1040144 1040289|1130103       [3.0, 4.0]           [1.67, 1.45]         5.33         7.02           1.55       24.02        2             1
```

## Additional notes

The class name `TravelTimeMatrixDecomposed` is provisional per the discussion in #543 — happy to rename (e.g. `ExpandedTravelTimeMatrix`, matching r5r's nomenclature) if you prefer. As with the issue, this was written with the help of Claude Code; the code has been human-reviewed and validated locally against R5 v7.6.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


christophfink:
Hi Taylor! 

Thanks for your contribution. I am sorry this is taking so long on our side. I want to stress that your contribution is definitely welcome, I just haven’t had the time to even read through your proposal, let alone review the merge request. Please bear with me, and thanks for the patience!

Chris

christophfink:
A few quick comments: 

1. I think it would feel most natural to go along r5r’s nomenclature, so if you could go ahead an rename the classes to the according names, that would be fantastic. 
2. Also, please check the lines that are now not covered by tests and add appropriate unit tests (then the checks will run again and most likely all pass)
3. Thanks so much for using r5py and contributing to it! very much appreciated!

